Issue #3, opened 2020-08-13
Tower Backup and Restore roles


Tompage1994:

Tompage1994:
closed via #7

PR #7, opened 2020-08-29
Add backup, restore, and restructure collection

sean-m-sullivan:
### What does this PR do?
Changes names to be in line with other repositories to remove tower from role names. 
Adds a pretasks role so shared tasks can be used by install, backup, restore, and future roles. 
Adds Backup, Restore, and Kerberos roles. 

against a running tower, has been tested with Tower 3.6 and 3.7

#3


kedark3 (approved):

roles/install/meta/main.yml
```diff
@@ -0,0 +1,23 @@
...
+      versions:
+        - 7
+
+  galaxy_tags:
+    - linux
+    - system
+    - ansible
+
+dependencies:
+  - redhat_cop.tower_utilities.pre_tasks
```
Tompage1994, line 23: I really like this. I didn't know you could add roles here to make magic happen. I'll be remembering that for the future.

roles/kerberos/README.md
```diff
@@ -0,0 +1,85 @@
...
+krb_admin_hostname=redhatdomainctl
+```
+
+## License
+
+MIT
+
+## Author Information
+
+Tom Page
```
Tompage1994: You can take my name off this. I didn't contribute this in the first place. Either put yourself, or if you don't want to lay claim then you could just put redhat-cop.

roles/pre_tasks/README.md
```diff
@@ -0,0 +1,86 @@
+# redhat_cop.tower_utilities.pre_tasks
```
Tompage1994, line 1: You may perhaps in here state explicitly that this role is pulled in by other roles in the meta dependencies otherwise you may end up with people trying to call it directly before calling the install role.

roles/pre_tasks/README.md
```diff
@@ -0,0 +1,86 @@
...
+# Playbook to install Ansible Tower as a single node
+
+- name: Install Ansible Tower
+  hosts: tower
+  become: true
+  vars:
+    tower_tower_releases_url: https://releases.ansible.com/ansible-tower/setup-bundle
+    tower_tower_release_version: bundle-3.6.3-1.tar.gz
+  roles:
+    - ansible-tower-install
```
Tompage1994: `- redhat_cop.tower_utilities.install`

roles/pre_tasks/README.md
```diff
@@ -0,0 +1,86 @@
...
+- name: Setup Ansible Tower
+  hosts: localhost
+  become: true
+  vars:
+    tower_hosts:
+      - "clusternode[1:3].example.com"
+    tower_database: "dbnode.example.com"
+    tower_database_port: "5432"
+  roles:
+    - ansible-tower-install
```
Tompage1994: `- redhat_cop.tower_utilities.install`

roles/restore/tasks/restore.yml
```diff
@@ -0,0 +1,10 @@
+---
+# Run the Setup to restore tower
+- name: "[Tower] Run the Ansible Tower Setup Program with restore option"
+  become: true
+  command: ./setup.sh -e 'restore_backup_file={{ restore_location | quote }}' -r
```
Tompage1994, line 5: I think this should have a default of 

```
restore_location: "{{ tower_working_location }}/tower-backup-latest.tar.gz"
```

or perhaps be even better than that and have:
```
restore_file: "tower-backup-latest.tar.gz"
restore_location: "{{ tower_working_location }}/{{ restore_file }}"
```

Tompage1994 (changes requested):
Looking good. I've got a few changes for you in there but the general transition is good. Seems to be no breaking changes to before other than rename of `tower_install` to `install`.

Once these fixes are implemented we should do a `v0.2.0` release and update `tower_configuration_examples` with the changes and create a separate playbook example for backup/restore.

kedark3:
I'd request @sean-m-sullivan @Tompage1994 to take a look at https://github.com/kedark3/ATCasC/tree/master/roles/tower-backup which is variation of backups I had written earlier this year. It was working fine for us. We were using just JSON dump backup as well.

roles/install/meta/main.yml
sean-m-sullivan: Yes its great, its also possible to chain roles this way where you only call the last one in the dependency list.

Tompage1994:
> I'd request @sean-m-sullivan @Tompage1994 to take a look at https://github.com/kedark3/ATCasC/tree/master/roles/tower-backup which is variation of backups I had written earlier this year. It was working fine for us. We were using just JSON dump backup as well.

That looks good. The full backup uses the same idea as this one. The meta backup is using the `tower_recieve` module which is deprecated in favour of `tower_export`. I imagine that will instead be a separate role though

kedark3:
@Tompage1994 yup you're right. just wanted to put it on our radar that we can support such meta backups/restores!

sean-m-sullivan:
> > I'd request @sean-m-sullivan @Tompage1994 to take a look at https://github.com/kedark3/ATCasC/tree/master/roles/tower-backup which is variation of backups I had written earlier this year. It was working fine for us. We were using just JSON dump backup as well.
> 
> That looks good. The full backup uses the same idea as this one. The meta backup is using the `tower_recieve` module which is deprecated in favour of `tower_export`. I imagine that will instead be a separate role though

We could likely build it into this role, using the tower export, then user can choose which kind of backup they want, same with restore. If user wants both they'd call the role twice with two different variable sets. Will look over this some more and see about integrating Kedar's work.

Tompage1994:
> > > I'd request @sean-m-sullivan @Tompage1994 to take a look at https://github.com/kedark3/ATCasC/tree/master/roles/tower-backup which is variation of backups I had written earlier this year. It was working fine for us. We were using just JSON dump backup as well.
> > 
> > 
> > That looks good. The full backup uses the same idea as this one. The meta backup is using the `tower_recieve` module which is deprecated in favour of `tower_export`. I imagine that will instead be a separate role though
> 
> We could likely build it into this role, using the tower export, then user can choose which kind of backup they want, same with restore. If user wants both they'd call the role twice with two different variable sets. Will look over this some more and see about integrating Kedar's work.

I think my personal preference would be separate roles to do things in different ways (i.e. probably export and import roles). Also I think smaller PRs are better, should probably get this one finished before moving onto the next thing? (again just personal preference)

roles/restore/tasks/restore.yml
sean-m-sullivan: Updated

roles/pre_tasks/README.md
sean-m-sullivan: Updated

roles/pre_tasks/README.md
sean-m-sullivan: Updated

roles/pre_tasks/README.md
sean-m-sullivan: Updated the description here, may need tweaking.

roles/kerberos/README.md
sean-m-sullivan: You wrote it more then I did, The backup/restore are very bare bones as well, its the pretasks that does most of our heavy lifting, and after that a wrapper on the installer, changed to rehat-cop

Tompage1994: Not the Kerberos role though!

roles/install/README.md
```diff
@@ -83,4 +83,4 @@ MIT
 
 ## Author Information
 
-Tom Page
+Redhat-Cop.
```
Tompage1994: Sorry, I meant change the author to not be me for the kerberos role. This role can stay as me as I did contribute it!

roles/pre_tasks/README.md
Tompage1994: Happy with that 👍

Tompage1994 (approved):
LGTM

Tompage1994 (changes requested):
Sorry, meant to click request changes on above review... If you just change back the author info in the install role and change it to redhat COP for the Kerberos role.

Tompage1994 (approved):

PR #9, opened 2020-09-10
Tower install ocp

### LINKED ISSUE #6: Tower OpenShift Install
We want to be able to install Tower on OpenShift in the same way we can install Tower normally. 

A `tower_install_ocp` (or similar) role should be created which follows the same pattern of `tower_install`

crenwick93:
### What does this PR do?
Adds a role to deploy a containerised Ansible Tower instance with optional external database.

Please follow the README in regards to running the new role. If successful the playbook will run through and you should see Ansible tower in Openshift.

closes #6 

N/A


roles/install_ocp/.travis.yml
```diff
@@ -0,0 +1,29 @@
+---
```
Tompage1994: Can you delete this file please

roles/install_ocp/README.md
```diff
@@ -0,0 +1,121 @@
+# ansible-tower-install
```
Tompage1994: should be `redhat_cop.tower_utilities.install_ocp`

roles/install_ocp/handlers/main.yml
```diff
@@ -0,0 +1,2 @@
+---
```
Tompage1994: You can delete this file

roles/install_ocp/tests/inventory
```diff
@@ -0,0 +1,2 @@
+localhost
```
Tompage1994: delete the `/tests/` directory

roles/install_ocp/meta/main.yml
```diff
@@ -0,0 +1,23 @@
...
+      versions:
+        - 7
+
+  galaxy_tags:
+    - linux
+    - system
+    - ansible
+
+dependencies:
+  - redhat_cop.tower_utilities.pre_tasks
```
Tompage1994: Can you edit this so it is 
```
dependencies:
  - role: redhat_cop.tower_utilities.pre_tasks
    vars:
      tower_releases_url: "{{ tower_releases_url }}"
      tower_setup_file: "{{ tower_setup_file }}"
```
I think that'll work.

Reason being that variable precedence will be off as role vars take precedence over play vars. Instead by adding this as a role param it will be correct (assuming the parsing for this works correctly). Can you then move those two vars from `vars` to `defaults`

roles/install_ocp/vars/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+# vars file for install_ocp
+tower_releases_url: https://releases.ansible.com/ansible-tower/setup_openshift/
+tower_setup_file: ansible-tower-openshift-setup-{{ tower_release_version }}.tar.gz
```
Tompage1994: Please move these two to `defaults` as per comment in `meta.yml`

roles/install_ocp/defaults/main.yml
```diff
@@ -0,0 +1,42 @@
...
+# and a service account must be created for auto peer-discovery to work.
+
+openshift_host: https://openshift.example.com
+openshift_skip_tls_verify: true
+openshift_project: tower
+openshift_user: kubeadmin
+
+# persistant volume for pg database details
+openshift_pg_pvc_name: postgresql
+pvc_claim_size: 10gi
```
Tompage1994: This failed for me in openshift 3.11. It needs to be `10Gi`.

Tompage1994 (changes requested):
Really good stuff. There are a few comments to make changes but in general looks good. I have confirmed it working in 3.11 and I think you've done it in 4.x? So we should be good.

There are a load of linting errors so if you could also clean those up and make the changes then we should be good to go.

crenwick93:
Updated the code to reflect those changes, should be good to go now :)

roles/install_ocp/meta/main.yml
```diff
@@ -0,0 +1,26 @@
...
+
+  galaxy_tags:
+    - linux
+    - system
+    - ansible
+
+dependencies:
+  - role: redhat_cop.tower_utilities.pre_tasks
+    vars:
+      tower_releases_url: "{{ tower_releases_url }}"
```
Tompage1994: So this doesnt seem to work (my bad).

Probably the best thing to do is change the vars name here to `tower_ocp_releases_url` and `tower_ocp_setup_file` then have 

```
dependencies:
  - role: redhat_cop.tower_utilities.pre_tasks
    vars:
      tower_releases_url: "{{ tower_ocp_releases_url }}"
      tower_setup_file: "{{ tower_ocp_setup_file }}"
```

Tompage1994 (changes requested):
Almost there. Just one broken thing

roles/install_ocp/defaults/main.yml
```diff
@@ -0,0 +1,46 @@
...
+
+############################################################
+#               Tower Installation Vars                    #
+############################################################
+
+# Working location for installation files
+tower_working_location: "/var/tmp"
+
+# vars file for install_ocp
+tower_releases_url: https://releases.ansible.com/ansible-tower/setup_openshift/
```
Tompage1994:
```
tower_ocp_releases_url
tower_ocp_setup_file
```

Tompage1994 (changes requested):

Tompage1994 (approved):
LGTM 

Thanks for the contribution @crenwick93

roles/install_ocp/README.md
```diff
@@ -0,0 +1,121 @@
...
+# OCP Requirements
+OpenShift 3.11+
+
+Per pod default resource requirements:
+* 6GB RAM,
+* 3CPU cores
+
+A setup and running Openshift cluster
+
+Admin privileges for the account running the openshift installer (cluster-admin role is required)
```
kedark3, line 24: What operations require the `cluster-admin` role?

roles/install_ocp/README.md
```diff
@@ -0,0 +1,121 @@
...
+pg_sslmode: 'prefer'  # set to 'verify-full' for client-side enforced SSL
+
+# Note: The user running this installer will need cluster-admin privileges.
+# Tower's job execution container requires running in privileged mode,
+# and a service account must be created for auto peer-discovery to work.
+
+# Deploy into Openshift
+# =====================
+
+openshift_host: https://openshift.example.com
```
kedark3, line 71: Is there a way to inject CA certs into the tower deployment? Like if I were to add RH-IT-Root-CA cert? How would we achieve that? Once pods are created, I believe user won't have sufficient permissions to add new certs and `update-ca-trust`, which I have experienced once. Hence wondering if we can add that somehow in initial deploy?

kedark3 (commented):
Looks great, can't wait to test drive this!!!

roles/install_ocp/README.md
Tompage1994: I think these requirements come directly from the tower docs 

https://docs.ansible.com/ansible-tower/3.7.2/html/administration/openshift_configuration.html

roles/install_ocp/README.md
Tompage1994: Surely this is all done at the openshift route level?

roles/install_ocp/README.md
```diff
@@ -0,0 +1,123 @@
...
+---
+# Playbook to install Ansible Tower as a single node
+
+- name: Install Ansible Tower on OCP
+  hosts: localhost
+  become: true
+  vars:
+    tower_release_version: 3.7.2-1
+    tower_releases_url: https://releases.ansible.com/ansible-tower/setup_openshift/
+    tower_setup_file: ansible-tower-openshift-setup-{{ tower_release_version }}.tar.gz
```
sean-m-sullivan: I updated the vars, I think we need to rework our pretasks to take into account openshift, and/or even bring this into the fold of the openshift installer. Instead of two different roles. However this is great work, straightforward, I think we need more documenation about what is needed but that will be provided in our examples. Great role addition.

sean-m-sullivan (approved):
Looks good, I made once change in the readme vars, as I found it installing regular tower instead of openshift tower. I did run into an issue with the installer, but thats the installers issue, this role preformed everything perfectly when testing on my home system.

roles/install_ocp/README.md
sean-m-sullivan: @Tompage1994 going to have you review this change before we merge, just in case you think we shouldn't add it. Otherwise I think we are good for merge.

Tompage1994: Ah I've realised why this is needed. I know Chris put in a change to negate the need for this but looks like he did it wrong and I didnt test it properly. I'll request a change where it's needed

roles/install_ocp/meta/main.yml
```diff
@@ -0,0 +1,26 @@
...
+
+  galaxy_tags:
+    - linux
+    - system
+    - ansible
+
+dependencies:
+  - role: redhat_cop.tower_utilities.pre_tasks
+    vars:
+      tower_ocp_releases_url: "{{ tower_releases_url }}"
```
Tompage1994: @crenwick93 youve done this the wrong way around

It should read 
`tower_releases_url: "{{ tower_ocp_releases_url }}"`

And similar for the other var here

Tompage1994 (changes requested):
One comment, then this is all good to go

roles/install_ocp/meta/main.yml
sean-m-sullivan: I've tested it with 
```
tower_releases_url: "{{ tower_ocp_releases_url }}"
tower_setup_file: "{{ tower_ocp_setup_file }}"
```
And it worked without issues.

crenwick93: Thanks, just pushed that change :)

Tompage1994 (approved):

roles/install_ocp/README.md
kedark3: Not really sure. Haven't gotten to try out this role yet. :(

PR #10, opened 2020-10-14
Add connection vars to inventory template and readme

sean-m-sullivan:
### What does this PR do?
Allows user to set connection vars in the inventory file. Would also allow for other custom vars. Not sure if there is a way to do this better, as some are mutually exclusive, and may use other options.

Will upload an example of using this to tower examples but has been tested.


roles/pre_tasks/templates/inventory.j2
```diff
@@ -9,6 +9,12 @@
 [all:vars]
 admin_password='{{ tower_admin_password }}'
 
+{% if tower_ssh_connection_vars is defined %}
+{% for item in tower_ssh_connection_vars %}
+{{ item.name}}='{{ item.value }}'
```
Tompage1994, line 14: This is presumably a lint error. The linting failed anyway but looks like for another reason.

Tompage1994 (changes requested):
Theres a linting error (Or at least the linting failed for whatever reason). I've marked one piece that looks wrong.

Could you give it a look over please.

Otherwise looks good to me

roles/pre_tasks/templates/inventory.j2
sean-m-sullivan: @Tompage1994 I've been hunting down trying to figure this one out, if anyone has any ideas I'd be grateful.

sean-m-sullivan: @Tompage1994 Tracked it down, wasn't a linting issue, it was a build issue in our linting precheck system, switched over to the precommit system we use in tower_config and added a few issues, Not sure on role names, but I believe since our galaxy.yml is variablized it doesn't pick up the role names correctly, Also on other tasks it expects chmod to be set, we will need to test that, but created an issue for it.

Tompage1994 (approved):

PR #15, opened 2020-10-28
Switch sorting of rabbitmq comparaison and add variable to defaults

ericzolf:
### What does this PR do?

Switch sorting of rabbitmq comparaison and add variable to defaults
    
the rabbitmq_required was comparing first with a version number then with 'latest', but 'latest'
can't be compared with a version, so it would fail before going to the 2nd step. Turning this
around solved the issue.
I also added the 'tower_release_version' variable to the defaults, where it was missing. 'latest' is the default, hope this is OK, for most demo/learning purposes, it's good enough I think.

And while I was at it, I've add Vim swap files to gitignore (bad habit I know).

Install Tower with the 'latest' version resp. a fixed version and see that it doesn't fail any more.

n/a

n/a

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ collections/*
 !collections/requirements.yml
 galaxy.yml
 *.tar.gz
+
+# Ignore Vim working files
+.*.swp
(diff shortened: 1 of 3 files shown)


Tompage1994 (approved):
Looks good to me. Good catch on the latest causing a bug.

Everything else seems sensible 👍

PR #17, opened 2020-10-30
Isolated nodes install

crenwick93:
### What does this PR do?
Adds compatibility for the automated configuring of isolated node groups

I have tested this manually. The first test was adding one isolated group with one isolated nodes. The second test was two isolated groups, each with a single node underneath them. Both tests passed. Attached are screen shots of the playbook completing (inventory populating correctly and tower installing), as well as Ansible Tower with the new instance groups.

N/A

This PR is based off the manual process of adding isolated nodes to an inventory file. More information here: https://developers.redhat.com/blog/2017/12/20/understanding-ansible-tower-isolated-nodes/ 

image

image

### DIFF
--- a/roles/install/README.md
+++ b/roles/install/README.md
@@ -44,7 +44,11 @@ tower_ssh_connection_vars: ''
 isolated_groups:
   - name: dmz1
     hostnames:
-    - isolatednode.example.com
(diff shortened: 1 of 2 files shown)


Tompage1994 (approved):
Looks good to me

PR #18, opened 2020-11-02
Fix README for install

Tompage1994:
### What does this PR do?
Fixes the README for the install role which had the wrong role name in the example playbook

N/A

N/A

N/A


ericzolf:
Looked good to me.

PR #21, opened 2020-11-27
Add RPM pre-requisites to pre_tasks and simplify role

ericzolf:
### What does this PR do?

On a minimal RHEL 8, the unarchive task complains about
`Commands "gtar" and "tar" not found. Command "unzip" not found.`
Also there is no point in having a tasks file beside main, also removed
unneeded prefix to task names.

Install Tower on RHEL8 with only the following packages and see how it complains without this PR:

```
%packages
@^minimal-environment
kexec-tools
%end
```

Then see that it works with the patch.

n/a

n/a


sean-m-sullivan (approved):
Looks good.

PR #23, opened 2020-12-09
Improve virtualenv role with better variables and more

ericzolf:
- all variables are properly documented and named properly in defaults/main.yml
- prefix of variable is always tower_venv
- support of RHEL8 added for RPM installation
- backup couldn't have worked, better naming also
- some re-shuffling to avoid code duplication, especially regarding directory creation

I didn't test the offline installation part, but it shouldn't be working worse than it used to work.


sean-m-sullivan (approved):
tested and looks good.

PR #24, opened 2020-12-15
Kerberos Update

sean-m-sullivan:
### What does this PR do?
Updates to the documentation to make them more in line with the collection
Updates to add missing kerberos parameter
Removed file that was leftover from previous role
Tested vs Windows server 2019. 

This should be merged to the said branch, then that branch can be merged to devel. 

Already tested vs windows server 2019


roles/kerberos/README.md
```diff
@@ -9,6 +12,7 @@ An Ansible role to configure Ansible Tower nodes to manage Windows systems.
 |`krb_realms`|yes|N/A|List of Dictionaries|Used for storing the realm(domain) name and its domain controllers.  A single domain controller can be specified if that is the only available option, but more than one is preferable. See example usage below.|
 |`krb_default_realm`|no|undefined|string|If a default realm(domain) is required to be specified this can be set.  Otherwise it remains unset in the krb5.conf.|
 |`krb_dns_lookup_realm`|yes|"false"|string|Whether or not to lookup DNS via realm.|
+|`dns_lookup_kdc`|no|"true"|string|Indicate whether DNS SRV records should be used to locate the KDCs and other servers for a realm.|
```
ericzolf: The variable is correctly called `krb_dns_lookup_kdc`.

ericzolf (changes requested):
Looks good to me even if I can't test it, with one documentation glitch. A description of a test setup on one's laptop would be way cool (could also be useful for other occasions like preparing for a Windows gig).

Tompage1994 (approved):
LGTM other than the one change already pointed out

roles/kerberos/README.md
sean-m-sullivan: I've updated the documentation, and added a file about setting up a windows AD domain. Had referred to my old Lab notes from College and put them in a markdown.

PR #25, opened 2020-12-15
update tower install

willtome:
### What does this PR do?
Update Tower Install role

```
- hosts: localhost
   roles:
     - redhat_cop.tower_utilities.install
```

Integration into redhat-cop/agnosticD and ansible/workshops

### DIFF
--- a/roles/install/tasks/tower_install.yml
+++ b/roles/install/tasks/tower_install.yml
@@ -21,7 +21,7 @@
       async: 10000
       poll: 20
       changed_when: false   # these will always run and will always report “changed” otherwise
-    
+
     - name: Wait for ansible tower to be running.
       uri:
         url: "{{ tower_url }}"
--- a/roles/install/tasks/tower_install.yml
+++ b/roles/install/tasks/tower_install.yml
@@ -1,21 +1,36 @@
 ---
-# Run the Setup
-- name: "[Tower] Run the Ansible Tower Setup Program"
-  become: true
-  command: ./setup.sh
-  args:
-    chdir: "{{ tower_setup_dir }}"
-  async: 10000
-  poll: 20
-  changed_when: false   # these will always run and will always report “changed” otherwise
-
-- name: Wait for ansible tower to be running.
+- name: Check Ansible Tower Running
   uri:
-    url: "{{ tower_server }}"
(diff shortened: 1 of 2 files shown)


sean-m-sullivan:
I'm not against these changes, but the check for tower release version is in pre-tasks, and is a meta dependency of this role
https://github.com/redhat-cop/tower_utilities/blob/devel/roles/pre_tasks/tasks/main.yml

Also not sure on the removal of tower alive after check, could see about reducing the time, but would want to test it.

willtome:
@sean-m-sullivan this change checks to see if Tower is already installed and skips the installer if it is. On the note of pre_tasks role, why does that exist separately to begin with?

roles/install/tasks/tower_install.yml
```diff
@@ -1,21 +1,36 @@
...
+    - name: Wait for ansible tower to be running.
+      uri:
+        url: "{{ tower_url }}"
+        status_code: 200
+        validate_certs: false
+      register: result
+      until: result.status == 200
+      retries: 90
+      delay: 10
+  when:
```
sean-m-sullivan: There is a trailing whitespace here that needs fixed. line 34 after the when

sean-m-sullivan:
pre_tasks exists separately as its common tasks used by install, install_ocp, restore, and backop.

But realize that now on the change, Should be fine once the whitespace issue fixed.

ericzolf:
Sorry if I reacted too late but I am the strong opinion that this MR was wrong: in the previous state, it was possible to use the role to fix, extend, upgrade Tower, and now it can only be used to initially install Tower. I still see the revert button and am very tempted to press it, but of course we should first discuss this.

Tompage1994:
> Sorry if I reacted too late but I am the strong opinion that this MR was wrong: in the previous state, it was possible to use the role to fix, extend, upgrade Tower, and now it can only be used to initially install Tower. I still see the revert button and am very tempted to press it, but of course we should first discuss this.

I think I agree with you here. Maybe instead we could have some variable which will force the installer to run (perhaps by default)

sean-m-sullivan:
I think we should just do another PR to fix it, I've had bad experiences with the revert before.

Issue #2, opened 2020-08-13
Add tower-install & tower-certs role


Tompage1994:
Remove from tower-configuration-examples and update that repo to make use of this collection

Issue #1, opened 2020-08-13
Set up as a collection


Tompage1994:
Set up ansible collection, with github actions to push to galaxy as per tower-configuration

PR #26, opened 2020-12-17
Updated to remove become and a python env

sean-m-sullivan:
### What does this PR do?
After doing some testing in an Github runner that will be brought here soon, I found a few problems with the role.
#1 the become requirement wasn't needed and could be removed. 
#2 the users ansible.cfg should be used to set the python environment if it is not set, not the inventory file as inventory takes precedent over config. 

Has been tested on openshift 4.6.


Tompage1994 (approved):
LGTM. Do we definitely want `-vvv`? Would it be better to have this as a variable?

sean-m-sullivan:
We can change that to a variable, let me look at that.

sean-m-sullivan:
Updated to have option for verbose install, and tested.

PR #28, opened 2020-12-18
Update process

sean-m-sullivan:
### What does this PR do?
After running into issue where I had no sudo access, created a test that only runs the yum if the package is not present.

Has been tested


ericzolf:
Not going yet into the details but does it make sense? It adds quite a lot of complexity and you can't anyway install Tower without root/sudo access, so what would be the point of running pre_tasks?

sean-m-sullivan:
Part of this is the changes to the tower_configuration I put forward, At least for Openshift/AWX on kubernetes, you don't need sudo access to install tower if the packages are preinstalled. I've been pulling this collection from galaxy to do the build, but my personal branch.

roles/pre_tasks/README.md
```diff
@@ -13,6 +13,9 @@ Available variables are listed below, along with default values defined (see def
 ```yaml
 tower_working_location: "/root/"
 
+# Use the default tower installation template
+pre_tasks_process_template: True
```
ericzolf: We should use `true` without capital because it's the only variant that will be accepted in YAML 1.2 (which will come, exactly like IPv6).

roles/pre_tasks/tasks/main.yml
```diff
@@ -1,10 +1,23 @@
 ---
 
 # Install absolutely necessary packages
+
+- name: Check if the absolutely necessary packages are installed
+  shell: "type {{ item }}"
```
ericzolf: `/usr/bin/type` is also a thing, so you don't need the shell and can use `command`.

roles/pre_tasks/tasks/main.yml
```diff
@@ -1,10 +1,23 @@
...
+  register: type_packages
+  ignore_errors: yes
+
 - name: "Install absolutely necessary packages"
   yum:
-    name: "{{ tower_pre_tasks_packages }}"
-    state: present
+    name: "{{ package_item.item }}"
+    state: latest
+  ignore_errors: true
```
ericzolf: you shouldn't need to ignore errors here? You want to fail if a required package can't be installed.

roles/pre_tasks/tasks/main.yml
ericzolf: I'm also not very happy about using `type` because it relies on the fact that the package is named like the command it contains, that could change and makes the code confusing. I would rather suggest `rpm -q {{ item }}`.

roles/pre_tasks/tasks/main.yml
```diff
@@ -1,10 +1,23 @@
 ---
 
 # Install absolutely necessary packages
+
+- name: Check if the absolutely necessary packages are installed
+  shell: "type {{ item }}"
+  with_items: "{{ tower_pre_tasks_packages }}"
+  register: type_packages
+  ignore_errors: yes
```
ericzolf, line 8: you should add a `changed_when: false` as the command doesn't change anything.

roles/pre_tasks/tasks/main.yml
```diff
@@ -1,10 +1,23 @@
...
 - name: "Install absolutely necessary packages"
   yum:
-    name: "{{ tower_pre_tasks_packages }}"
-    state: present
+    name: "{{ package_item.item }}"
+    state: latest
+  ignore_errors: true
+  loop: "{{ type_packages.results }}"
+  loop_control:
+    loop_var: package_item
```
ericzolf, line 17: To be consistent, you should also rename the loop_var on the previous loop, or rename none of them.

roles/pre_tasks/tasks/main.yml
```diff
@@ -1,10 +1,23 @@
...
+  with_items: "{{ tower_pre_tasks_packages }}"
+  register: type_packages
+  ignore_errors: yes
+
 - name: "Install absolutely necessary packages"
   yum:
-    name: "{{ tower_pre_tasks_packages }}"
-    state: present
+    name: "{{ package_item.item }}"
+    state: latest
```
ericzolf: `[E403] Package installs should not use latest` -> replace with `present`

roles/pre_tasks/tasks/main.yml
ericzolf: Thinking about it, you could even avoid a loop by doing something like `rpm -q {{ tower_pre_tasks_packages | join(' ') }}` and then in the next loop, change the condition to `when: ('package ' + item + ' is not installed') in __towerutils_check_packages.stdout`.

e.g. the output of `rpm -q unzip untar 2>/dev/null` is:
```
unzip-6.0-49.fc33.x86_64
package untar is not installed
```

roles/pre_tasks/tasks/main.yml
```diff
@@ -1,10 +1,23 @@
 ---
 
 # Install absolutely necessary packages
+
+- name: Check if the absolutely necessary packages are installed
+  shell: "type {{ item }}"
+  with_items: "{{ tower_pre_tasks_packages }}"
+  register: type_packages
```
ericzolf: A bit unfair I know but while I'm at it: following the [best practices from oasis](https://github.com/oasis-roles/meta_standards/blob/master/README.md), the variable should be called something like `__towerutils_check_packages`.

ericzolf (changes requested):

ericzolf:
Sorry in the delay for reviewing, and happy new year.

roles/pre_tasks/tasks/main.yml
sean-m-sullivan: Updated, Agree with changes

roles/pre_tasks/tasks/main.yml
sean-m-sullivan: Updated, Agree with changes

roles/pre_tasks/tasks/main.yml
sean-m-sullivan: Updated, Agree with changes, updated this var like oasis

roles/pre_tasks/tasks/main.yml
sean-m-sullivan: Updated, Agree with changes

roles/pre_tasks/README.md
sean-m-sullivan: Updated, Agree with changes

roles/pre_tasks/tasks/main.yml
sean-m-sullivan: I'm trying to remember why I did this, I think because it failed when you don't have root access, Will do testing to see about removing

roles/pre_tasks/tasks/main.yml
sean-m-sullivan: Will do testing around this, Don't disagree with it, was more trying to find a way to verify that things were installed when you don't have root access.

roles/pre_tasks/tasks/main.yml
sean-m-sullivan: Updated to remove, the when should handle this case

roles/pre_tasks/tasks/main.yml
sean-m-sullivan: Got this working in most cases, but when testing here: https://github.com/sean-m-sullivan/tower_configuration/runs/1691976044

It skips the task somehow and then errors out because its not skipping the install absolutely necessary packages, Hoping to utilize this process for tower on both tower_utilties and tower_configuration builds so we can eval vs awx and tower.

sean-m-sullivan: Updated, and tested, Looks like good changes!

ericzolf (approved):

PR #30, opened 2021-01-18
Introduce basic installation of Automation Hub

ericzolf:
### What does this PR do?
the basic configuration works, using the database deployed by the Tower installer, and self-signed certificates;
a more complex configuration with standalone/external database should be possible, but not tested. I didn't even look at the SSL configuration aspects.
I took also the chance to rename 'tower_database' to 'tower_database_host' which is clearer.

Add the `tower_ah_...` variables to your inventory and re-run the setup.

n/a

The installation fails when installing a single node Tower with embedded database, and with the AH using the embedded database, because the setup playbooks don't open the firewall. I only added a commented out potential fix, subject to discussion (because @sean-m-sullivan said he needs the role to work without admin rights).

### DIFF
--- a/roles/pre_tasks/README.md
+++ b/roles/pre_tasks/README.md
@@ -38,7 +38,7 @@ tower_setup_file: ansible-tower-setup-{{ tower_release_version }}.tar.gz
 tower_hosts:
   - "localhost ansible_connection=local"
 
-tower_database: ""
+tower_database_host: ""
 tower_database_port: ""
 
 tower_ssh_connection_vars: ''
@@ -92,7 +92,7 @@ $ ansible-playbook playbook.yml -e @tower_vars.yml tower
   vars:
     tower_hosts:
       - "clusternode[1:3].example.com"
-    tower_database: "dbnode.example.com"
+    tower_database_host: "dbnode.example.com"
     tower_database_port: "5432"
   roles:
     - redhat_cop.tower_utilities.install
--- a/roles/pre_tasks/README.md
+++ b/roles/pre_tasks/README.md
@@ -42,7 +42,25 @@ tower_database_host: ""
 tower_database_port: ""
 
 tower_ssh_connection_vars: ''
+
+# as long as the host name is empty, Automation Hub will NOT be installed
+tower_ah_hosts: []
+
+# the default configures an AH using the default database with similar defaults
+# as Tower
+tower_ah_admin_password: "{{ tower_admin_password }}"
+# if tower_database_host isn't defined or is empty,
+# the host where the installation takes place is deemed to host the DB
+tower_ah_pg_host: "{{ tower_database_host | default(ansible_host | default(inventory_hostname), true) }}"
+tower_ah_pg_port: "{{ tower_database_port }}"
+tower_ah_pg_database: 'automationhub'  # it is NOT supported to use the same name as for Tower!
+tower_ah_pg_username: "{{ tower_pg_username }}"
+tower_ah_pg_password: "{{ tower_pg_password }}"
+tower_ah_pg_sslmode: prefer
 ```
+
+> **CAUTION:** even if it is a list, it hasn't been tested that the installation of more than one AH is possible,
+	especially as it would surely require a mean to have multiple database names.
 
 ## tower_ssh_connection_vars
 
--- a/roles/pre_tasks/tasks/main.yml
+++ b/roles/pre_tasks/tasks/main.yml
@@ -49,3 +49,16 @@
     src: inventory.j2
     dest: "{{ tower_setup_dir }}/inventory"
   when: pre_tasks_process_template
+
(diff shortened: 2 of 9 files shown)


roles/pre_tasks/tasks/main.yml
```diff
@@ -49,3 +49,16 @@
     src: inventory.j2
     dest: "{{ tower_setup_dir }}/inventory"
   when: pre_tasks_process_template
+
+# FIXME when installing Tower standalone with the embedded database (i.e. without adding a host to the [database] group) and configuring AH to use the same DB, it fails because the firewall isn't opened for PostgreSQL by the installation playbooks. Not sure yet if it's something we should fix in this role, like:
```
ericzolf, line 53: This is the part I hesitated to add without prior discussion...

sean-m-sullivan: Going to do some testing, It worries me if this is not in the default install for AH, but more, should it be included with the regular installer, but will test this in my home lab.

IDEA: include this, but its disabled by default, and if user sets configure_firewall_external_database, then this task fires.

Tompage1994 (changes requested):
Looks good in general.

Could you add the new vars to the relevant READMEs please

ericzolf:
Addressed the documentation topic, but I left the firewall topic open for now.

Tompage1994 (approved):
Good with me now

roles/pre_tasks/README.md
```diff
@@ -38,12 +38,30 @@ tower_setup_file: ansible-tower-setup-{{ tower_release_version }}.tar.gz
 tower_hosts:
   - "localhost ansible_connection=local"
 
-tower_database: ""
+tower_database_host: ""
```
sean-m-sullivan, line 41: So I noticed an issue, We have these variables in both pre_tasks and in install, and in install_cop because they are all interconnected. While its fine to just change the vars in pretasks, the Readme's need to be updated in all 3 places so we can keep things straight. Is there a better way to do this? or should we just link to the pre tasks readme in the other 2 places? Not sure on what's the best answer, leaning towards keeping it in all 3, but only slightly. What are other peoples thoughts?

roles/pre_tasks/tasks/main.yml
sean-m-sullivan: So After testing, I was able to install on a fresh rhel machines with no firewall configured and access AH without issue. 1 server is the AH, one server the database.

sean-m-sullivan:
Once the Readme choice is solved this is solid to good.

roles/pre_tasks/README.md
ericzolf: I have the impression that it's only an impression :wink: 

The following command shows that `tower_database[...whatever...]` is documented almost everywhere, but only used in "pre_tasks":

```
tower_cert/README.md
backup/README.md
install/README.md
pre_tasks/defaults/main.yml
pre_tasks/tasks/main.yml
pre_tasks/templates/inventory.j2
pre_tasks/README.md
restore/README.md
```

I strongly suspect a wild cut&paste effect. We should do the same exercise of cleaning up README.md resp. `defaults/main.yml` as I did for the tower_configuration.

Nevertheless, I had overseen a few places to renamed `tower_database` to `tower_database_host` but I'll create a separate issue for the clean-up, if it's OK for you.

roles/pre_tasks/tasks/main.yml
ericzolf: Did you explicitly set the database host in the `[database]` group? This is what I didn't do, and most probably why it failed.

roles/pre_tasks/README.md
Tompage1994: It may be useful creating a doc file of the shared variables and importing them into each readme to save duplication. Not if you can do an include in markdown. Could be worth switching the readme to asciidoc as you can definitely do imports in that and it will still render in GitHub

ericzolf: What about having the discussion in the issue #32 I created so that this PR can be merged?

sean-m-sullivan: Yep we'll do it there, and merge.

roles/pre_tasks/tasks/main.yml
sean-m-sullivan: yes I set the database in the database group

sean-m-sullivan: Note, the url will not work I removed private keys and it only works for 30 min at the moment
```
- name: Install Ansible Tower
  hosts: localhost
  become: true
  vars:
    tower_releases_url: https://access.cdn.redhat.com/content/origin/files/
    tower_setup_file: ansible-automation-platform-setup-bundle-1.2.1-1.tar.gz
    tower_database_host: 192.168.122.120
    tower_hosts: []
    tower_ah_hosts:
      - 192.168.122.200
  roles:
    - redhat_cop.tower_utilities.install
```

sean-m-sullivan (commented):
Reviewed this again looks good.

PR #35, opened 2021-02-12
adding ... to the end of each yml file and added check in yamllint

djdanielsson:
### What does this PR do?
adds ... to the end of each yml file. Also added .gitattributes file to override linguist to calculate yml files as a language so it stops calling it html

automated

no

none


sean-m-sullivan (approved):
LGTM Thanks for the contribution!

PR #36, opened 2021-02-24
update docs

crenwick93:
### What does this PR do?
Removes the need for root access on the "control node/bastion" from which the playbook is run. (Recent customer has this issue). This fix will only run as root on the targeted machine not on the local machine. Most customers have root logins disabled.

This has been manually tested. to test just run the install as normal, remembering to specify ssh_connection_vars as described in the README"

N/A

N/A

### DIFF
--- a/roles/install/README.md
+++ b/roles/install/README.md
@@ -55,7 +55,7 @@ isolated_groups:
 
 ## tower_ssh_connection_vars
 
-Connection vars can be set in the inventory file through a list of vars. Please note, at minimum ansible_become should be set to true here as the Tower installation requires root access.
+Connection vars can be set in the inventory file through a list of vars.
 
 ```yaml
 tower_ssh_connection_vars:
--- a/roles/install/README.md
+++ b/roles/install/README.md
@@ -55,7 +55,7 @@ isolated_groups:
 
 ## tower_ssh_connection_vars
 
-connection vars can be set in the inventory file through a list of vars
+Connection vars can be set in the inventory file through a list of vars. Please note, at minimum ansible_become should be set to true here as the Tower installation requires root access.
 
 ```yaml
 tower_ssh_connection_vars:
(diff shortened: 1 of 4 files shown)


roles/install/README.md
```diff
@@ -67,6 +67,8 @@ tower_ssh_connection_vars:
     value: vagrant
   - name: ansible_ssh_private_key_file
     value: /path/to/file
+  - name: ansible_become
```
Tompage1994, line 70: Can you make the same change on `backup/README.md` as well please

roles/install/README.md
```diff
@@ -55,7 +55,7 @@ isolated_groups:
 
 ## tower_ssh_connection_vars
 
-connection vars can be set in the inventory file through a list of vars
+Connection vars can be set in the inventory file through a list of vars. Please note, at minimum ansible_become should be set to true here as the Tower installation requires root access.
```
Tompage1994: Not 100% sure that this is true in ALL cases so just remove this.

Tompage1994 (changes requested):
Looks good. Just a couple changes

Tompage1994 (approved):
LGTM

roles/install/tasks/tower_install.yml
```diff
@@ -15,7 +15,6 @@
 - block:
     # Run the Setup
     - name: "[Tower] Run the Ansible Tower Setup Program"
-      become: true
```
Tompage1994, line 18: @sean-m-sullivan just want to check with you this is a correct change to make. I can't think of a reason this should actually be set to true here. As pointed out by Chris, this shouldn't need to be true, and in some cases will not work if you don't have root access on the local machine (when using a bastion for example). In the situation you actually do want to run this as root locally the you'd run the whole thing as become I guess.

sean-m-sullivan: that I'm not sure, I'm going to do some testing on this branch just to verify. I know I came accross the root problem this week when standing up a new one.

sean-m-sullivan (approved):
LGTM

PR #45, opened 2021-05-21
Ah def

Tompage1994:
### What does this PR do?
Adds a sensible default to `tower_ah_url`

Run through deploying a ah only deployment without `tower_ah_url` will break previously. Now will not.

N/A

N/A


sean-m-sullivan (approved):
LGTM

PR #46, opened 2021-07-06
Add a check mode option to go a bit further in check mode

ericzolf:
### What does this PR do?
Just adding a `check_mode: false` to the RPM check to make the check mode go a little bit further without error (it helps making sure nothing is fully wrong).

Just call a playbook where you install Tower in check mode and see that it fails a bit later than without the patch...

n/a

n/a


Tompage1994 (approved):
LGTM

PR #47, opened 2021-07-06
Rename all cluster variables from hosts to nodes

ericzolf:
### What does this PR do?
It appears that tower_hosts clashes with tower_collection/roles/hosts' variable of same name;
and the other nodes (AH and database) for being consistent.

Inventory variables must be renamed and provisioning tried.

see Google Chat...

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,10 @@ collections:
 
 ## Release and Upgrade Notes
 
+- After v0.4.0 the variable `tower_hosts` was renamed to `tower_nodes` in order to not collide with the variable of the same name in the redhat\_cop.tower\_configuration collection.
+Similarly, the variables tower\_ah\_nodes, tower\_database\_node and tower\_ah\_pg\_node were introduced/renamed.
+Don't forget to adapt your inventory/ies and variable definitions accordingly.
+
 ## Roadmap
 
 ## Contributing to this collection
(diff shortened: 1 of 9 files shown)


Tompage1994 (approved):
I'm happy with this as a change. We need to be careful that the change in variable name is explained (possibly add something to the readme?).

Other than that I'm happy. Will wait for @sean-m-sullivan to approve also

README.md
```diff
@@ -38,6 +38,10 @@ collections:
 
 ## Release and Upgrade Notes
 
+- After v0.4.0 the variable `tower_hosts` was renamed to `tower_nodes` in order to not collide with the variable of the same name in the redhat\_cop.tower\_configuration collection.
```
Tompage1994, line 41: `- After v1.0.0 the variable `tower_hosts` was renamed to `tower_nodes` in order to not collide with the variable of the same name in the redhat\_cop.tower\_configuration collection.`

Tompage1994 (changes requested):
We are on v1.0.0 right now so if you could change the readme that would be great

sean-m-sullivan (approved):
Looks good to me, like the idea of nodes and haven't thought of it before. This will need rewritten to work with the new installer as well when it becomes official, I believe I have the inventory file for it though.

README.md
sean-m-sullivan: This need the version changed still to 1.0.0

Tompage1994 (approved):

PR #50, opened 2021-10-04
updating group name to automationcontroller for AAP2

### LINKED ISSUE #49: AAP 2 renamed groups
[tower] is now [automationcontroller] with the current AAP installer and it will error because it is using the old name.

djdanielsson:
### What does this PR do?
updated group name from tower to automationcontroller to work with AAP2

Automated

resolves #49 

none

### DIFF
--- a/roles/pre_tasks/defaults/main.yml
+++ b/roles/pre_tasks/defaults/main.yml
@@ -5,8 +5,8 @@
 #               Tower Installation Vars                    #
 ############################################################
 
-# If you are running AAP2.X+ use automationcontroller any earlier version use tower
-tower_or_aap: automationcontroller
+# Change to True if you are installing AAP1.X or Tower 3.X
+legacy_tower: False
 
 # Tower release version, either fixed e.g. 3.7.2 or "latest"
 tower_release_version: latest
--- a/roles/pre_tasks/defaults/main.yml
+++ b/roles/pre_tasks/defaults/main.yml
@@ -5,6 +5,9 @@
 #               Tower Installation Vars                    #
 ############################################################
 
+# If you are running AAP2.X+ use automationcontroller any earlier version use tower
+tower_or_aap: automationcontroller
+
 # Tower release version, either fixed e.g. 3.7.2 or "latest"
 tower_release_version: latest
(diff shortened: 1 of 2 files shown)


djdanielsson:
If you want to keep support for the older version then maybe we need to add a new variable to say something like aap2: true and in the template we do if aap2 automationcontroller else tower

Tompage1994 (approved):
Happy for this to go in as-is. 

When we get around to it we should probably rename the repo to aap_utilities or similar and release a new collection

@sean-m-sullivan just to check you're happy for this to go in and then once some work has gone into this we can roll it out as new.

sean-m-sullivan:
I'm ok with the change, but thinking we should variablize it. Either Tower or AAP, and keeping this that does the install for all supported 3.8 + 4.0+, but that might get complicated, or just be a different template. Thoughts?

Tompage1994:
Agreed, for now at least, let's get a variable (e.g. `legacy_tower`?)  which will make the template resolve correctly for tower/controller as relevant

djdanielsson:
was this more what you were looking for then?

roles/pre_tasks/defaults/main.yml
```diff
@@ -5,6 +5,9 @@
 #               Tower Installation Vars                    #
 ############################################################
 
+# If you are running AAP2.X+ use automationcontroller any earlier version use tower
+tower_or_aap: automationcontroller
```
Tompage1994: Sorry to be a pain, I think I would prefer it as a true/false value.

Perhaps `legacy_tower` and set the default to false. Then in the template you can just use some logic to set either [tower] or [automationcontroller].

The reason behind that is that it will just be easier for the end user (especially given there are only two possible values)

Tompage1994 (changes requested):

roles/pre_tasks/defaults/main.yml
djdanielsson: no problem

Tompage1994 (approved):

Tompage1994:
Thanks @djdanielsson

PR #52, opened 2021-10-07
fixing default tower_ah_url

### LINKED ISSUE #51: tower_ah_url only referenced in defaults as localhost
https://github.com/redhat-cop/tower_utilities/blob/dfe8f4a316f7dfc2fe339f41c42dcc949ec8bbe5/roles/install/tasks/tower_install.yml#L51

install fails because this is trying to hit localhost when checking if AH is up and running and not the host that AH is installed on. This value is set in defaults here
https://github.com/redhat-cop/tower_utilities/blob/dfe8f4a316f7dfc2fe339f41c42dcc949ec8bbe5/roles/install/defaults/main.yml#L9

djdanielsson:
### What does this PR do?
fix an issue where tower_ah_url would default to localhost and not the ah_node[0] as expected

Automated

resolves #51 

none

### DIFF
--- a/roles/install/defaults/main.yml
+++ b/roles/install/defaults/main.yml
@@ -6,7 +6,7 @@
 ############################################################
 
 tower_url: "https://localhost"
-tower_ah_url: "https://localhost"
+tower_ah_url: "https://{{ tower_ah_nodes[0] | default('localhost') }}"
 tower_server: "{{ tower_url }}"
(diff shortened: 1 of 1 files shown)


sean-m-sullivan (approved):
I like it, user can still override it if they want since its a default, but this is a much better flexible option.

PR #53, opened 2021-10-08
this was added to inventory in 2.0.1

djdanielsson:
### What does this PR do?
added some new variables that were added in 2.0.1

Automated

none

none


Tompage1994 (changes requested):
I suppose we should also allow the setting of registry_url as well in case they want to use an internal registry for images.

Also, I can't find these options documented anywhere, and don't know what they would do since I thought the whole container processing was manual at the moment?

Could you also just make sure the variables used here are defaulted.

Thanks

djdanielsson:
How is this?

sean-m-sullivan (approved):
This looks good in terms of defaulting, it matches the original inventory provided with the installer.

PR #54, opened 2021-12-04
add aap-setup-download role to download automatically the AAP setup file

ericzolf:
### What does this PR do?

Adds a role to automatically download the installer image from RHN

My playbook looks so far like this:

```yaml
- name: download and install AAP from the bastion
  hosts: bastion.*
  gather_facts: false
  become: true
  tags: aap_installation
  roles:
    - redhat_cop.tower_utilities.aap_setup_download
  post_tasks:
    - debug:
        var: aap_setup_down_installer_file
```
And the only variable I really need to define in my vault is aap_setup_down_offline_token (see `defaults/main.yml` for the details and other variables).

n/a

Adapted from https://www.redhat.com/en/blog/automating-installation-ansible-automation-platform-ansible-and-satellite

I still need to work on the "soft" aspects of the role, means documentation :-) hence still WiP, else it works good enough for me.


ericzolf:
Ready for review, I'm working currently on the follow-up roles `aap_setup_prepare` and `aap_setup_install` but it takes longer to properly test...

sean-m-sullivan:
I like the idea of this, and want to run it through some testing. However I think long term we need to incorporate the 
https://galaxy.ansible.com/middleware_automation/redhat_csp_download
plugin and/or role. As that does a lot of our lifting in a reproducible way. 

And looking at it, it might not be possible, BUT its something we should push for it to be available from if its not.

roles/aap_setup_download/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+You will need a Red Hat Ansible Automation Platform (AAP, hence the name) subscription.
+Once this is a given, you will be able to create yourself an offline token
+
+Role Variables
+--------------
+
+The following input variables are required:
+
+* `aap_setup_down_offline_token` contains your offline token as described in the requirements.
```
sean-m-sullivan, line 19: we should point the user in the right place to get the token, rather then going to the blog post. 
https://access.redhat.com/management/api is the link to follow.

roles/aap_setup_download/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+  uri:
+    url: "{{ aap_setup_down_token_url }}"
+    method: POST
+    body_format: form-urlencoded
+    body:
+      grant_type: refresh_token
+      client_id: rhsm-api
+      refresh_token: "{{ aap_setup_down_offline_token }}"
+  register: __aap_setup_down_login
+- name: collecting the available installers
```
sean-m-sullivan, line 14: Should be a line between this task and the previous task.

sean-m-sullivan (changes requested):
This works well, And I left the previous comment about the module and looked into it, its not listed on that release page, and this eliminates the need to look up and ID number that module would require, even if it were on that page. 
With the documentation change and the one lint/style change I think this would be good to go!

sean-m-sullivan:
I was trying out a new task line to get the file name, and then went back and decided to not do that. Either way there is a PR for the changes I mentioned.

PR #59, opened 2021-12-20
Add aap_setup_prepare role

ericzolf:
### What does this PR do?
Add aap_setup_prepare role

install AAP 2.1...

n/a

It works so far (I successfully installed a complete cluster on this basis) but there is still quite a lot to do:

1. the inventory-template is written in hard and needs to become a real template
2. the SSO and Catalog workers functionality hasn't been considered at all (I would probably see it anyway in a follow-up PR)


roles/aap_setup_prepare/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
+---
+# tasks file for aap_setup_prepare
+
+# we check first so that the installation can happen without root access
+- name: check if the absolutely necessary packages are installed
+  raw: "rpm -q {{ __aap_setup_prep_required_packages | join(' ') }}"  # noqa 303
```
djdanielsson, line 6: I am curious why do this and not just run the yum without the when?

roles/aap_setup_prepare/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+  register: __aap_setup_prep_packages_result
+  ignore_errors: yes
+  changed_when: false
+  check_mode: false  # run this command even in check mode
+
+- name: install absolutely necessary packages
+  yum:
```
djdanielsson, line 13: should we change this to dnf since AAP 2.1 requires RHEL8+

roles/aap_setup_prepare/tasks/main.yml
sean-m-sullivan: It wasn't working with the yum module before without ROOT access, this was the workaround. It allows the collection to be executed without root if the prereqs are there.

roles/aap_setup_prepare/tasks/main.yml
sean-m-sullivan: we should likely do that, though, I don't think it matters ansible wise, under the hood its likely still dnf.

djdanielsson: yes I agree it doesn't change much other than it would prevent it from working on RHEL 7

roles/aap_setup_prepare/tasks/main.yml
djdanielsson: but isn't root required to install AAP?

ericzolf: We're just talking about the use case where you start the cluster installation from a host _outside_ of the cluster, something like a bastion host which you might have to share with others and might not have root access.

ericzolf: You definitely need the use case if you create the VMs of the cluster within your automation (outside of the scope of this collection obviously, but feasible).

ericzolf:
It isn't really ready for proper review but I need to go and I don't want to block anybody willing to review and merge, it should work as it is, even if it isn't nice (but I can improve further next year).

As discussed per chat, the inventory-template is now just two big loops going over variables (one for the hosts, two for variables, secret or not-secret). This allows to maintain the group and variables only in `defaults/main.yml` and keep the template nice and tidy.

As an example, my inventory looks like this (don't do this in production with the ansible_password all over the place):

```yaml
---
aap_setup_prep_inv_nodes:
  automationcontroller:
    autoctl1.ewl.example.com: "ansible_user={{ ansible_user }} ansible_password={{ ansible_password }} ansible_become_password={{ ansible_become_password }} ansible_become=True"
    autoctl2.ewl.example.com: "ansible_user={{ ansible_user }} ansible_password={{ ansible_password }} ansible_become_password={{ ansible_become_password }} ansible_become=True"
  database:
    autodb.ewl.example.com: "ansible_user={{ ansible_user }} ansible_password={{ ansible_password }} ansible_become_password={{ ansible_become_password }} ansible_become=True"
  automationhub:
    autohub.ewl.example.com: "ansible_user={{ ansible_user }} ansible_password={{ ansible_password }} ansible_become_password={{ ansible_become_password }} ansible_become=True"

aap_setup_prep_inv_vars:
  all:
    # workaround to avoid enabling the codeready repo on the autohub server
    pulp_rhel_codeready_repo: []

    ## CONTROLLER PARAMETERS ##

    admin_password: '{{ ansible_password }}'

    ## CONTROLLER DATABASE PARAMETERS ##

    pg_password: '{{ ansible_password }}'
    pg_host: 'autodb.ewl.example.com'
    pg_port: '5432'
    pg_database: 'awx'
    pg_username: 'awx'
    pg_sslmode: 'prefer'  # set to 'verify-full' for client-side enforced SSL

    ## AUTOMATION HUB PARAMETERS ##

    automationhub_admin_password: '{{ ansible_password }}'

    ## AUTOMATION HUB DATABASE PARAMETERS ##

    automationhub_pg_host: 'autodb.ewl.example.com'
    automationhub_pg_port: '5432'
    automationhub_pg_database: 'automationhub'
    automationhub_pg_username: 'automationhub'
    automationhub_pg_password: '{{ ansible_password }}'
    automationhub_pg_sslmode: 'prefer'
```

And my vault looks like this:

```yaml
---
aap_setup_prep_inv_secrets:
  all:
    registry_username: "123456789|aap21-myuser"
    registry_password: "someverylongstringverysecret"
```

And the resulting inventory looks like:

```ini
[automationcontroller]
autoctl1.ewl.example.com ansible_user=myuser ansible_password=redhat ansible_become_password=redhat ansible_become=True
autoctl2.ewl.example.com ansible_user=myuser ansible_password=redhat ansible_become_password=redhat ansible_become=True

[database]
autodb.ewl.example.com ansible_user=myuser ansible_password=redhat ansible_become_password=redhat ansible_become=True

[automationhub]
autohub.ewl.example.com ansible_user=myuser ansible_password=redhat ansible_become_password=redhat ansible_become=True

[all:vars]
pulp_rhel_codeready_repo=[]
admin_password=redhat
pg_password=redhat
pg_host=autodb.ewl.example.com
pg_port=5432
pg_database=awx
pg_username=awx
pg_sslmode=prefer
automationhub_admin_password=redhat
automationhub_pg_host=autodb.ewl.example.com
automationhub_pg_port=5432
automationhub_pg_database=automationhub
automationhub_pg_username=automationhub
automationhub_pg_password=redhat
automationhub_pg_sslmode=prefer
registry_username=123456789|aap21-myuser
registry_password=someverylongstringverysecret
```
By lack of time, I didn't test it till the end, but looks OK to me :smiley:

sean-m-sullivan:
This looks good going to do some fiddiling or other things, I was going to test to see if the setup.sh can just take a yaml file like our tower config that the user would setup. It should work for everything but the groups/hosts. Going to do some testing, but Almost think thats the better option :)

sean-m-sullivan:
I just tested it, ./setup.sh -e @vars.yml does work if we wanted to go that route

ericzolf:
> This looks good going to do some fiddiling or other things, I was going to test to see if the setup.sh can just take a yaml file like our tower config that the user would setup. It should work for everything but the groups/hosts. Going to do some testing, but Almost think thats the better option :)

The issue with YAML is that it's more difficult to generate by template due to the changing indentations. Also, `prepare` is split from `install` exactly for people being able to push their own inventory between the two steps, if they want.

sean-m-sullivan:
had not considered that, was testing things to see what worked, and throwing things at the wall to see what sticks/works. At this point agree with the separation of prepare/install. I am almost thinking that this is something that might be easier solved by making improvements to the installer, to allow variable insertion, or kicking off the playbook itself. Still noodling around though on what would be best.

sean-m-sullivan:
Not sure when the edit comment was made, but I like the looks of it. I think its a LOT more maintainable, we should like have an example with things preset, but that's personal choice, and makes it so if more variables are introduced the 'install' role doesn't need to be updated.

roles/aap_setup_prepare/templates/inventory.j2
```diff
@@ -0,0 +1,19 @@
+{# output all the hosts and their groups #}
+
+{% for group_key in aap_setup_prep_inv_nodes %}
+[{{ group_key }}]
+{% for node_key in aap_setup_prep_inv_nodes[group_key] %}
+{{ node_key }} {{ aap_setup_prep_inv_nodes[group_key][node_key] }}
```
sean-m-sullivan: This line needs tweaking to
```
{{ node_key }}{% if aap_setup_prep_inv_nodes[group_key][node_key] is defined %} {{ aap_setup_prep_inv_nodes[group_key][node_key] }}{% endif %}

```
Make sure to include the space after the if.
So that if a user defines it as such:
```
aap_setup_prep_inv_nodes:
  automationcontroller:
    controller.node: "servicescatalog_controller_name=something"
  database:
    database
  automationhub:
    ah.node
```
It doesn't complain about it not being there. 

Otherwise I get the error:
AnsibleUnicode object' has no attribute 'd'

sean-m-sullivan (changes requested):
Need that one change and this looks great.

roles/aap_setup_prepare/templates/inventory.j2
ericzolf: Done!

ericzolf:
Ready for merging!

sean-m-sullivan (approved):
LGTM we need to add some documentation but I can get on that.

PR #62, opened 2022-01-19
Fix different glitches in aap prep and inst

ericzolf:
### What does this PR do?
- add framework for (temporary) fixes like AAP-1413
- quote ports explicitly to workaround AAP-1798
- replace raw by command for rpm call for security reasons
- make the definition of ah_hostname and controller_hostname truly optional

install end to end

see above

see above


sean-m-sullivan (approved):
Looks good, like the way to incorporate a fix.

PR #63, opened 2022-01-19
Prepare renaming of the collection to aap_utilities

ericzolf:
### What does this PR do?
- removed all obsolete roles
- corrected all ansible-lint issues
- renamed backup, restore and certs
- adapt aap_certs to cover also automation hub

```yaml
---
- name: assign given certificates
  hosts: aap21_servers
  gather_facts: false
  become: false
  roles:
    - redhat_cop.tower_utilities.aap_certs
```

with inventory containing something like:

```
aap_certs_controller_ssl_cert: "{{ playbook_dir }}/tower.cert"
aap_certs_controller_ssl_key: "{{ playbook_dir }}/tower.key"
aap_certs_autohub_ssl_cert: "{{ playbook_dir }}/pulp.cert"
aap_certs_autohub_ssl_key: "{{ playbook_dir }}/pulp.key"
aap_certs_create_backup: true
```

By lack of time, I didn't test the backup and restore roles, but they shouldn't be worse than they are today.

no

We already agreed to rename the collection to aap_utilities in the near future.

I didn't touch kerberos because I wasn't sure it's still relevant (we'd need it inside an EE probably).


sean-m-sullivan:
I kept looking for the pre task rename, forgetting this was already done. I think the only question is should we append the aap name to all of these when we name it aap_utils
It will come out as 
redhat_cop.aap_utils.aap_install

sean-m-sullivan:
After discussion it was 3 votes to strip the aap_ prefix from roles and make them still 
redhat_cop.aap_utils.install

ericzolf:
As per discussion in the Automation CoP meeting, we wait for a decision on the overall rule and then we act accordingly.

sean-m-sullivan (approved):
Approved, with the  name change

PR #64, opened 2022-01-28
Remove standards and point to good practices

ericzolf:
### What does this PR do?
Remove standards file and point to good practices, so that it's aligned without need to maintain anything.


djdanielsson:
sounds like a good idea to me

PR #75, opened 2022-06-24
Add markdown linter

branic:
### What does this PR do?

This PR adds the markdownlint linter to be run as part of the GitHub Actions. In addition it fixes all of the linter errors in the various markdown files.

Run the GitHub action for push/PR and see that the markdown linter step runs

No

None


.markdownlint.yml
```diff
@@ -0,0 +1,16 @@
+default: true
```
Tompage1994, line 2: Is there any way this file can be moved to `.github/workflow-config/.markdownlint.yml` and have an option added to the runner to load the config file from there in the same way we have done for ansible-lint?

galaxy.yml.j2
```diff
@@ -13,15 +13,18 @@ repository: {{ collection_repo }}
...
     - galaxy.yml.j2
-    - release.yml
     - .github
+    - .markdownlint.yml
+    - .gitignore
+    - .gitattributes
 license:
     - GPL-3.0-or-later
 tags:
-    - tower
```
Tompage1994, line 21: I would advocate for leaving the `tower` tags in here as this collection can (I think) still be used to deploy Tower and certainly the older versions can be.

Tompage1994 (changes requested):
Mostly looks good. Just a couple of small changes

Tompage1994 (changes requested):
I've just taken a look and it seems there are other implementations of markdownlint which can be integrated into our current pre-commit hooks. 

I've been through and done the same thing for ah_configuration so if you could do something similar. See here: https://github.com/redhat-cop/ah_configuration/compare/devel...Tompage1994:mdlint

.markdownlint.yml
branic: I can, but the advantage of having the config file in the root folder is that users using linters in an IDE (such as VSCode) will use the config when editing files. If the config file is in a different directory it isn't picked up by the IDE linters by default. I was going to propose that the other linter configs be moved to the root folder for this reason. 

IMHO it is a better workflow for the user to be able to see linter errors in the IDE before pushing a commit as opposed to getting the errors after pushing. Another argument for this setup is that GitHub Actions is limited to displaying 10 annotations for each step and 50 per 50 (https://github.community/t/annotation-limitation/17998/2), so if there are more than 10 linter errors the user won't be able to see them unless they look at the details of the action run.

galaxy.yml.j2
branic: That makes sense to me. I'll add Tower back in.

.markdownlint.yml
Tompage1994: I kind of get that argument. We originally moved it all out of the root of the repo and we have tried to make everything come under the pre-commit. I think this is another tool we should bring into the pre-commit. Definitely understand where you're coming from in terms of IDE integration and maybe that's something we need to think about.

@sean-m-sullivan what do you think?

branic: Bringing it into pre-commit I don't think is a problem. Pre-commit and the tools being run can use config files in the root folder. In fact the "Hack" that is in place to copy the pre-commit config from the workflow-config folder to the root folder could be eliminated as well if the config files were move to the root folder. The config files should be added to the galaxy.xml.j2 so they are not included in the collection build.

sean-m-sullivan: I'm back and forth on this, We definitely want to add it to the Galaxy build ignore files, but with IDE things like this reading automatically, and some of the linters wanting it in base, it might be time to move it to the TLD and leave it there, we initially moved it to the sub folders just to keep things 'clean', and I think the disadvantages are starting to outweigh the advantage of that.

sean-m-sullivan: TLD, my vote is to move it to base, as its giving more advantages then it did 2 years ago.

roles/aap_backup/README.md
```diff
@@ -102,7 +102,7 @@ tower_ssh_connection_vars:
 The following playbook and accompanying vars file containing the defined seed objects can be invoked in the following manner.
 
 ```sh
-$ ansible-playbook playbook.yml -e @tower_vars.yml tower
+ansible-playbook playbook.yml -e @tower_vars.yml tower
```
sean-m-sullivan: lets change this to controller as well.

sean-m-sullivan (changes requested):
Other then the one change about the tower var file, this looks good.

.markdownlint.yml
branic: :+1: Sounds good. I'll move them to TLD.

roles/aap_backup/README.md
branic: Changed

.markdownlint.yml
branic: I changed the markdownlint to use the pre-commit action and moved all the linter configs to TLD. I was also able to remove the separate yamllint and just let ansible-lint perform the yamllint as it now picks up the .yamllint.yml which extends its configuration. Some lint errors have also been fixed.

sean-m-sullivan: @Tompage1994 Handing this over to you for final approval, I think its at a good place, but with the TLD change, and everything going on, want you to at least have a chance to review before we merge.

sean-m-sullivan (approved):
LGTM

Tompage1994 (approved):
Looks good. I'm assuming this has run through properly and everything fixed? If not we can just go back and do that after. 

Thanks for the work on this!

PR #77, opened 2022-06-27
Allow RHEL9 install

### LINKED ISSUE #76: RHEL9 support
while going through these roles I noticed that we have rhel8 hard coded here 

https://github.com/redhat-cop/aap_utilities/blob/0791918be8b50198acdfe0b488a55ea2af11c432/roles/aap_setup_download/vars/main.yml#L3 

I think we should change this to default 8 but if we gather facts and this is being run on a RHEL9 server it should pull the rhel9 copy.

Tompage1994:
### What does this PR do?

Allows download of the RHEL9 version of Controller

```
- hosts: localhost
  vars:
    aap_setup_down_offline_token: aaabbbccc0123
    aap_setup_down_version: "2.2"
    aap_setup_down_dest_dir: .
    aap_setup_down_type: setup
    aap_setup_rhel_version: "9"
  roles:
    - ../roles/aap_setup_download
```

resolves #76

### DIFF
--- a/roles/aap_setup_download/README.md
+++ b/roles/aap_setup_download/README.md
@@ -17,9 +17,11 @@ The following input variables are required:
 It has no default value and _must_ be defined.
 * `aap_setup_down_version` defines the minor version to download (e.g. `2.1`), make sure you defines it as string and not as float!
 The default is the latest version available at time of writing.
+The default is the latest version available at time of writing.
 * `aap_setup_down_dest_dir` is the directory to where you want to download the tarball.
 It is by default the working directory `aap_setup_working_dir` also used by other roles of the collection, or ultimately `/var/tmp`.
 * `aap_setup_down_type` can be either `setup` or `setup-bundle`, depending which flavour of the tarball you want to download.
+* `aap_setup_rhel_version` defines the major RHEL version being used (currently 8 or 9). If you are gathering facts you possibly don't need to specify this as the role will attempt to work out the value required though you will if AAP will be installed on machines on a different OS than the installer will run on.
 
 The full path of the downloaded file is stored in the fact `aap_setup_down_installer_file` so that it can be used for extraction.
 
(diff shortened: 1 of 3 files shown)


roles/aap_setup_download/README.md
```diff
@@ -17,9 +17,11 @@ The following input variables are required:
 It has no default value and _must_ be defined.
 * `aap_setup_down_version` defines the minor version to download (e.g. `2.1`), make sure you defines it as string and not as float!
 The default is the latest version available at time of writing.
+The default is the latest version available at time of writing.
```
djdanielsson: duplicate line

djdanielsson (changes requested):
Thanks for updating this so fast! I found one thing to change.

roles/aap_setup_download/defaults/main.yml
```diff
@@ -9,6 +9,9 @@
 # which version of the installer to download
 aap_setup_down_version: "2.1"
 
+# Which RHEL version are you using (8 or 9)
+aap_setup_rhel_version: "{{ ansible_distribution_major_version | default(8, true) }}"
```
branic, line 13: Should the default here be 9 instead of 8 (to match what the readme change states)?

djdanielsson: I thought of that when reading it, I was not sure if the meaning was latest for that version and the major version default would be 8 for now. I am not sure we should push 9 until at least 9.1 or 9.2.

Tompage1994: Yes the default should be 8 I think still. The readme doesn't specify which one is default but I can add that in

branic: I think leaving the default at 8 makes sense and with the change to the readme it is clearer now which one will be selected.

djdanielsson (approved):
LGTM

Issue #78, opened 2022-06-27
Change the Instance checkers to use the inventory variables


sean-m-sullivan:
https://github.com/redhat-cop/aap_utilities/blob/13f7f661c35cbc66c1b47f4f5c565381cd130332/roles/aap_setup_install/tasks/main.yml#L4-L32

For the status checks we should use the inventory variables
automationhub_admin_password
admin_password

sean-m-sullivan:
I accidentally bricked my install, and fixed this in the process.

djdanielsson:
so this is fixed? @sean-m-sullivan

PR #79, opened 2022-06-29
update hostnames to inventory names

sean-m-sullivan:
### What does this PR do?
This updates the hostnames to be pulled from the variables set for the install, rather then vars that are used no where else. 
It also sets the default download version to 2.2

It has been tested and works with the current deployment variables.

Fixes Issue #78

### DIFF
--- a/roles/aap_setup_download/defaults/main.yml
+++ b/roles/aap_setup_download/defaults/main.yml
@@ -7,7 +7,7 @@
 # aap_setup_down_offline_token:
 
 # which version of the installer to download
-aap_setup_down_version: "2.1"
+aap_setup_down_version: "2.2"
 
 # Which RHEL version are you using (8 or 9)
 aap_setup_rhel_version: "{{ ansible_distribution_major_version | default(8, true) }}"
--- a/roles/aap_setup_install/defaults/main.yml
+++ b/roles/aap_setup_install/defaults/main.yml
@@ -1,6 +1,10 @@
 ---
 # defaults file for aap_setup_install
 
+# Hostnames for check
+controller_hostname: aap_setup_prep_inv_nodes['automationcontroller'][0]
+ah_hostname: aap_setup_prep_inv_nodes['automationhub'][0]
+
 # where is the setup directory
 aap_setup_inst_setup_dir: "{{ aap_setup_prep_setup_dir }}"
 
--- a/roles/aap_setup_install/tasks/main.yml
+++ b/roles/aap_setup_install/tasks/main.yml
@@ -3,11 +3,11 @@
 
 - name: Check Ansible Tower Running
   uri:
-    url: "https://{{ controller_hostname }}/api/v2/ping/"
+    url: "https://{{ controller_hostname }}/"
     method: GET
-    user: "{{ controller_username }}"
-    password: "{{ controller_password }}"
-    validate_certs: "{{ controller_validate_certs | default(omit) }}"
+    user: admin
+    password: "{{ aap_setup_prep_inv_vars.all.admin_password }}"
+    validate_certs: "{{ controller_validate_certs | default('false') }}"
     force_basic_auth: true
   register: __aap_setup_inst_ctl_check
   ignore_errors: true
@@ -20,9 +20,9 @@
   uri:
     url: "https://{{ ah_hostname }}/api/galaxy/"
     method: GET
-    user: "{{ ah_username }}"
-    password: "{{ ah_password }}"
-    validate_certs: "{{ ah_validate_certs | default(omit) }}"
+    user: admin
+    password: "{{ aap_setup_prep_inv_vars.all.automationhub_admin_password }}"
+    validate_certs: "{{ controller_validate_certs | default('false') }}"
     force_basic_auth: true
   register: __aap_setup_inst_ah_check
   ignore_errors: true
@@ -43,7 +43,7 @@
 
     - name: wait for automation controller to be running
       uri:  # use the first host from the list if no hostname is defined
-        url: "https://{{ controller_hostname | default((aap_setup_prep_inv_nodes['automationcontroller'].keys() | list)[0]) }}"
+        url: "https://{{ controller_hostname }}/"
         status_code: 200
         validate_certs: "{{ controller_validate_certs | default(omit) }}"
       register: __aap_setup_inst_result
@@ -54,7 +54,7 @@
 
     - name: wait for automation hub to be running
       uri:  # use the first host from the list if no hostname is defined
-        url: "https://{{ ah_hostname | default((aap_setup_prep_inv_nodes['automationhub'].keys() | list)[0]) }}/ui/"
+        url: "https://{{ ah_hostname }}/api/galaxy/"
         status_code: 200
         validate_certs: "{{ ah_validate_certs | default(omit) }}"
       register: __aap_setup_inst_result_ah
--- a/roles/aap_setup_prepare/defaults/main.yml
+++ b/roles/aap_setup_prepare/defaults/main.yml
@@ -137,6 +137,5 @@ aap_setup_prep_inv_secrets: {}
 #     #isolated_key_generation: true
 
 # list of fixes to apply to the installation scripts, use with CAUTION!
-aap_setup_prep_fixes:
-  - aap_1413
+aap_setup_prep_fixes: []
 ...


roles/aap_setup_download/defaults/main.yml
```diff
@@ -7,7 +7,7 @@
 # aap_setup_down_offline_token:
 
 # which version of the installer to download
-aap_setup_down_version: "2.1"
+aap_setup_down_version: "2.2"
```
Tompage1994, line 10: Not a change, more a question... Do you think we could find some magic to work out what the latest version is and pull that?

roles/aap_setup_install/tasks/main.yml
```diff
@@ -3,11 +3,11 @@
 
 - name: Check Ansible Tower Running
   uri:
-    url: "https://{{ controller_hostname }}/api/v2/ping/"
+    url: "https://{{ controller_hostname | default((aap_setup_prep_inv_nodes['automationcontroller'].keys() | list)[0]) }}"
```
Tompage1994: Would it make sense to still have `/api/v2/ping/` at the end?

```suggestion
    url: "https://{{ controller_hostname | default((aap_setup_prep_inv_nodes['automationcontroller'].keys() | list)[0]) }}/api/v2/ping/"
```

roles/aap_setup_install/tasks/main.yml
```diff
@@ -3,11 +3,11 @@
...
     method: GET
-    user: "{{ controller_username }}"
-    password: "{{ controller_password }}"
-    validate_certs: "{{ controller_validate_certs | default(omit) }}"
+    user: admin
+    password: "{{ aap_setup_prep_inv_vars.all.admin_password }}"
+    validate_certs: "{{ controller_validate_certs | default('false') }}"
```
Tompage1994, line 10: Default behaviour here should be true I think?

roles/aap_setup_install/tasks/main.yml
```diff
@@ -18,11 +18,11 @@
 
 - name: Check Automation Hub Running
   uri:
-    url: "https://{{ ah_hostname }}/api/galaxy/"
+    url: "https://{{ ah_hostname | default((aap_setup_prep_inv_nodes['automationhub'].keys() | list)[0]) }}/ui/"
```
Tompage1994: Again, I think this should probably be the api endpoint?
```suggestion
    url: "https://{{ ah_hostname | default((aap_setup_prep_inv_nodes['automationhub'].keys() | list)[0]) }}/api/galaxy/"
```

roles/aap_setup_download/defaults/main.yml
djdanielsson: I like that idea but I'm not sure how we could do it, its frustrating that they create a new repo each release.

djdanielsson: ok I looked around a bit... it wouldn't be pretty but we could create a loop that checks what version returns items from the API and then pick the largest version number from the list to go with

sean-m-sullivan: I would be for SOMETHING, but as dj said, last I checked it was gnarly.

roles/aap_setup_install/tasks/main.yml
sean-m-sullivan: I was default making it work out of the box for anyone, if it were normal module, I'd say yes true.

roles/aap_setup_install/tasks/main.yml
sean-m-sullivan: Good catch

roles/aap_setup_download/defaults/main.yml
Tompage1994: Probably not something for this PR anyway. I had a poke around looking through the APIs and couldn't find anything nice

roles/aap_setup_install/tasks/main.yml
Tompage1994: ok, fair enough

roles/aap_setup_install/tasks/main.yml
sabre1041: I was about to create an issue, but isnt `aap_setup_prep_inv_nodes['automationcontroller'].keys() | list)[0]` incorrect? 

`aap_setup_prep_inv_nodes['automationcontroller']` is a list, so there are no keys

roles/aap_setup_install/tasks/main.yml
sean-m-sullivan: Fixed and updated

roles/aap_setup_install/tasks/main.yml
sean-m-sullivan: Andrew is right, fixed and updated.

sean-m-sullivan:
Turned off that aap fix as well, as it was causing issues.

roles/aap_setup_install/tasks/main.yml
```diff
@@ -18,11 +18,11 @@
 
 - name: Check Automation Hub Running
   uri:
-    url: "https://{{ ah_hostname }}/api/galaxy/"
+    url: "https://{{ ah_hostname | default(( aap_setup_prep_inv_nodes['automationhub'][0] ) }}/api/galaxy/"
```
Tompage1994: Does it not make sense to default these in the role defaults rather than here?

Tompage1994 (changes requested):
Just a change around where to keep the defaults then all good

roles/aap_setup_install/tasks/main.yml
ericzolf: Till now I had the impression that our approach was to do the right thing by default. Ignoring certification validation doesn't seem to me to be the "right thing" ™, or what do you think? And as our good practices state, we should define (explicit) defaults in the defaults/main.yml, not within the code!

roles/aap_setup_install/tasks/main.yml
ericzolf: I second this opinion, in accordance with our good practices. I'm also not sure anymore where all these variables not respecting the role's prefix are coming from (suspecting that they're historical). Some clean-up of all those variables might not hurt.

roles/aap_setup_install/tasks/main.yml
Tompage1994: I'm now a little on the fence on this. If the user doesn't specify their own certs on install then this will definitely be self signed and will definitely fail. Other thing we could look at doing is adding a bit of code to trust the CA that gets created in 2.2 then we can default 'true'. Otherwise probably best to just leave as false for this because it's likely the user will either later add their own certs or will trust the certs themselves (but they've just been generated so that can't have happened yet)

Tompage1994: (sorry for the mind dump) 🤮

djdanielsson: so we could do this with jinja and have it check if they pass a cert to default true else it will be false. I also like the idea of of adding the self signed to trust which is probably a better fix over all

Tompage1994: > so we could do this with jinja and have it check if they pass a cert to default true else it will be false. I also like the idea of of adding the self signed to trust which is probably a better fix over all

Only problem I can think with that is that you could be running this from anywhere(i.e. not RHEL) I usually run it from my Mac which means that probably wont be so easy

djdanielsson: I thought the validate certs stuff only was for the servers connections between themselves.... is that not correct?

Tompage1994: This task will be likely running from localhost (or a bastion) and therefore wont be in the cluster so it wouldn't necessarily be RHEL.

In 2.2 installer a common CA is created and all the hosts trust each other which is kind of cool and we could leverage that by delegating this command to the controller host - that could potentially be better functionality anyway?

djdanielsson: I thought the validate certs were to set settings in controller/hub/db type stuff and that's all so it doesn't matter where this is being run from. Is that not correct?

Tompage1994: No, if you look this task referenced here is just calling out to the controller API to see if it's up. It would be done from the target host this role is running against (often localhost)

roles/aap_setup_install/tasks/main.yml
sean-m-sullivan: I put it in defaults that will pick up. And the origin is indeed historical

roles/aap_setup_install/defaults/main.yml
```diff
@@ -1,6 +1,10 @@
 ---
 # defaults file for aap_setup_install
 
+# Hostnames for check
+aap_controller_hostname: "https://{{ controller_hostname | default(( aap_setup_prep_inv_nodes['automationcontroller'][0] ) }}"
```
Tompage1994: this is a net new variable isn't it? If so it needs to be documented. But I don't think its necessary? Can't we just have:

`controller_hostname: aap_setup_prep_inv_nodes['automationcontroller'][0]`

and the same thing with ah?

Tompage1994: And then in the tasks have the uri module start with:

```yaml
uri:
  url: "https://{{ controller_hostname }}/api/v2/ping/"
```

as before.

i.e. the change we want here is to default the controller_hostname var that we had before.

Tompage1994 (changes requested):

roles/aap_setup_install/defaults/main.yml
sean-m-sullivan: You are right, I'm trying to reinvent multiple wheels

sean-m-sullivan: Fixed!

djdanielsson (approved):
LGTM

Tompage1994 (approved):

PR #81, opened 2022-07-05
Add AAP-4555 setup.log fix to aap_setup_install

### LINKED ISSUE #80: Move away setup.log before restarting an installation
I realized that setup.log isn't overwritten by setup.sh if it already exists, so that the log gets lost in such cases (at least if running setup.sh as non-root). There should hence be the possibility to move away a potentially existing setup.log. It's not as easy as it sounds because Ansible doesn't have a good option to _move_ things (I don't think we want to simply _remove_ the log file).

ericzolf:
### What does this PR do?

setup.log is moved away before a new installation starts.
It's not by default, the new variable aap_setup_inst_fixes needs to contain aap_4555

Just start twice the same installation with pre-existing setup.log file and see how a new setup.log is created.

resolves #80 

https://issues.redhat.com/browse/AAP-4555

### DIFF
--- a/roles/aap_setup_install/defaults/main.yml
+++ b/roles/aap_setup_install/defaults/main.yml
@@ -63,7 +63,7 @@ aap_setup_inst_extra_vars: {}
 aap_setup_inst_force: false
 
 # list of fixes to apply before starting the actual installation,
-# empty by default, possible values are listed below
-aap_setup_inst_fixes: []
-# - aap_4555  # setup.log is only saved on first call as non-root
+# review before bumping up the default AAP version
+aap_setup_inst_fixes:
+  - aap_4555  # setup.log is only saved on first call as non-root (2.2.0)
 ...
--- a/roles/aap_setup_install/defaults/main.yml
+++ b/roles/aap_setup_install/defaults/main.yml
@@ -61,4 +61,9 @@ aap_setup_inst_extra_vars: {}
 
 # force the installation of AAP even if it's already running
 aap_setup_inst_force: false
+
+# list of fixes to apply before starting the actual installation,
+# empty by default, possible values are listed below
+aap_setup_inst_fixes: []
+# - aap_4555  # setup.log is only saved on first call as non-root
 ...
--- a/roles/aap_setup_install/tasks/fixes/aap_4555.yml
+++ b/roles/aap_setup_install/tasks/fixes/aap_4555.yml
@@ -6,6 +6,7 @@
   ansible.builtin.stat:
     path: "{{ aap_setup_inst_setup_dir }}/setup.log"
   register: __aap_setup_inst_log
+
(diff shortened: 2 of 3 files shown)


roles/aap_setup_install/defaults/main.yml
```diff
@@ -61,4 +61,9 @@ aap_setup_inst_extra_vars: {}
 
 # force the installation of AAP even if it's already running
 aap_setup_inst_force: false
+
+# list of fixes to apply before starting the actual installation,
+# empty by default, possible values are listed below
+aap_setup_inst_fixes: []
```
Tompage1994: Should we, by default, apply the fixes? I think it probably makes sense to

roles/aap_setup_install/tasks/fixes/aap_4555.yml
```diff
@@ -0,0 +1,21 @@
+---
+# Fix for https://issues.redhat.com/browse/AAP-4555 - it'll move away the
+# setup.log before starting the installation
+
+- name: AAP_4555 | check if setup.log does exist
+  ansible.builtin.stat:
+    path: "{{ aap_setup_inst_setup_dir }}/setup.log"
+  register: __aap_setup_inst_log
+- name: AAP_4555 | create a backup of setup.log if it exists
```
Tompage1994, line 10: Could you please put line breaks between each task. I'm not sure why the linter hasn't picked up on it here

Tompage1994: ok, apparently ansible-lint doesn't check for that. Who knew...?

Either way, if you could add it, it makes it look nicer :)

Tompage1994 (changes requested):
Couple of minor changes requested. Thanks for this workaround

roles/aap_setup_install/defaults/main.yml
ericzolf: Well, I enabled by default the last fix I had added to the prep role, and it broke as we moved to 2.2 so I thought that I'd make it not applied by default. I'm still hesitant to be honest and have no strong opinion. I'll add it, we should anyway review the fixes before bumping the version up.

roles/aap_setup_install/tasks/fixes/aap_4555.yml
ericzolf: I agree, my oversight.

Tompage1994 (approved):
Looks good.

RE: breaking when the issue is fixed, it shouldn't really be an issue as it won't actually break anything here. We should still review these fixes once a new version is released.

I wonder if there is some magic we could do to have fixes dependant on version so when we select what version to install we auto-select a fix. Perhaps there isn't real value in doing that though as the number of fixes we have is low.

PR #82, opened 2022-07-07
Update ansible-lint version

branic:
### What does this PR do?

Updates ansible-lint to the current version (6.3.0) and fixes linting errors detected by the newer version.

Run ansible-lint against code base.

No

None


sean-m-sullivan (approved):
LGTM, thank you for doing this work!! its much Appreciated!

I think we are going to need to do some testing on minimal version needed to use this with tower 3.8, as I think it still works, with the role versions, and if ansible even checks that in a collection. But I think thats an issue of itself, so approving.

PR #86, opened 2022-08-07
Added auth to setup validation

### LINKED ISSUE #85: Automation Hub Fails to Validate
The task "wait for automation hub to be running" fails to return a successful result (HTTP 403) as the Galaxy API requires authentication credentials be provided

sabre1041:
The `aap_setup_install` does not currently provide authentication during the validation phase after installation. Without providing credentials, validation never succeeds

Resolves #85

### DIFF
--- a/roles/aap_setup_install/tasks/main.yml
+++ b/roles/aap_setup_install/tasks/main.yml
@@ -50,6 +50,10 @@
     - name: wait for automation controller to be running
       ansible.builtin.uri:  # use the first host from the list if no hostname is defined
         url: "https://{{ controller_hostname }}/"
+        method: GET
+        user: admin
+        password: "{{ aap_setup_prep_inv_vars.all.admin_password }}"
+        force_basic_auth: true
         status_code: 200
         validate_certs: "{{ controller_validate_certs | default(omit) }}"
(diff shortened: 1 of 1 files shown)


djdanielsson (approved):
Over all this seems to be a fine change, though from my testing the controller should pass every time without username / pass but hub will fail. I am not sure why both do not point to either api endpoints or just the UI.

sabre1041:
@djdanielsson agree. @sean-m-sullivan might be able to provide some insights

branic:
IMHO both should be checking the UI url and not require login. The UI for Hub will return a 403 status until it is ready. The UI for controller will return a 200 status before it is ready and while migrations are running so the until should check for that in addition the the 200 status code like:

```
until: (__aap_setup_inst_result['status'] == 200) and ('migrations_notran' not in __aap_setup_inst_result['url'])
```

sean-m-sullivan:
The controller should not require login, and we should probably put in a when its defined in the groups as if someone uses this for JUST one or the other it will fail, However if the corresponding one is configured, it should have the password.However I do find one change needed.

sean-m-sullivan (approved):
NM, was thinking we needed no log, but the module should handle that.

PR #87, opened 2022-08-09
Add support for OCP based install

branic:
### What does this PR do?

This PR adds a role for installing AAP on OpenShift.

This is a fairly basic role at this point. I plan on adding support for installing/configuring RH-SSO integration with AH in future PR(s). 

I'd also like feedback/thoughts on whether the role should support creating multiple AC and/or AH instances (in one or more namespaces) before the PR is merged because the variable structure will need to change slightly (which would be a breaking change I'd like to avoid if possible).

Create a playbook to run the `aap_ocp_install` role with appropriate variables set (see the README.md in the role).

No

None


vvaldez:
Very nicely organized role, I will give it a try. Thanks!

sean-m-sullivan (commented):
I want to test this out, and double check before approving, but this looks awesome, I have a few initial comments.

galaxy.yml.j2
```diff
@@ -35,3 +35,5 @@ tags:
     - tower_utils
 dependencies:
     "ansible.posix": ">=1.0.0"
+    "kubernetes.core": ">=2.2.0"
```
sean-m-sullivan, line 38: The collection requirements and the role meta collections are different, I believe these are all 4 collections correct?

collections:
  - redhat.openshift
  - community.okd
  
  Just checking that we have the right dependencies declared.

roles/aap_ocp_install/templates/hub/instance.yaml.j2
```diff
@@ -0,0 +1,42 @@
+---
+apiVersion: automationhub.ansible.com/v1beta1
+kind: AutomationHub
+metadata:
+  name: {{ aap_ocp_install_hub['instance_name'] }}
+  namespace: {{ aap_ocp_install_namespace }}
+spec:
```
sean-m-sullivan, line 7: What do you think of templating out the spec, or allowing a user to provide their own for both installs.

djdanielsson:
I am curious why would someone use this over the operator in the store that can deploy AAP in OCP?

sean-m-sullivan:
I believe this Uses the operator in the store, but lets you Automate the process rather then click click click

galaxy.yml.j2
branic: community.okd is the upstream version of redhat.openshift. I chose to depend on redhat.openshift for the collection as that is the Red Hat supported collection. The role however will use either community.okd or redhat.openshift for the tasks. 

I'm thinking about opening an RFE for collection dependencies to introduce support for depending on one or another collection and have a preference for one over the other. There are several collections where this would be useful (awx.awx and ansible.controller is another example).

roles/aap_ocp_install/templates/hub/instance.yaml.j2
branic: That is one of the additional PRs I'm planning on doing. I didn't want to make this PR more complicated. I would like to introduce the ability to have many (if not all) parameters for the CR's templated and able to be set by the user through variables.

branic:
> I am curious why would someone use this over the operator in the store that can deploy AAP in OCP?

This role automates the deployment of the operator and AC/AH CRs that the operator will use to deploy the instances. The same question could be asked of the roles that support installing on bare metal/vms. Why wouldn't a user just use the setup.sh that is shipped with AAP.

sean-m-sullivan (approved):
Sorry this took to long, it kept getting pushed back in the to do list. I've been able to test it, and review it, and everything looks good. I look forward to the future increased templating, and this becoming popular :), this is really well done, I was going through with some comments for changes, but things were already in place.

Issue #90, opened 2022-08-30
controller_hostname and ah_hostname within aap_setup_install not variabilised


alawong:
within the aap_setup_install role, 2 default variables, controller_hostname, and ah_hostname are used to check if controller and AH are running. However, they are set to (literally):

controller_hostname: aap_setup_prep_inv_nodes['automationcontroller'][0]
ah_hostname: aap_setup_prep_inv_nodes['automationhub'][0]

As they are missing the brackets, these tasks fail.

Tompage1994:
Good spot. I've accepted your PR (just waiting for it to run through CI)

As a workaround before this can be released, you can set extra vars for the values which are broken, namely:

```
controller_hostname: "{{ aap_setup_prep_inv_nodes['automationcontroller'][0] }}"
ah_hostname: "{{ aap_setup_prep_inv_nodes['automationhub'][0] }}"
```

PR #91, opened 2022-08-30
Correct controller_hostname and ah_hostname within aap_setup_install …

alawong:
…defaults file

### What does this PR do?

Variabilised controller_hostname and ah_name within the defaults file of the aap_setup_install role. 

This should be tested by running a playbook that uses the role aap_setup_download (assuming the installer isn't already downloaded), aap_setup_prepare, and aap_setup_install on a server without AAP2 installed. The playbook should fully complete running, and the task 'wait for automation controller to be running' should return an 'ok' message.

Resolves 'controller_hostname and ah_hostname within aap_setup_install not variabilised' #90 

https://github.com/redhat-cop/aap_utilities/issues/90

N/A

### DIFF
--- a/roles/aap_setup_install/defaults/main.yml
+++ b/roles/aap_setup_install/defaults/main.yml
@@ -2,8 +2,8 @@
 # defaults file for aap_setup_install
 
 # Hostnames for check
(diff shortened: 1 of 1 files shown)


Tompage1994 (approved):
LGTM

Thanks for the contribution!

PR #92, opened 2022-08-30
Parametrize OCP controller install

branic:
### What does this PR do?

Adds paramatization to all settings for the controller installation on OCP via the operator. Also introduces the ability to have the controller installed in a different namespace than the operator (useful for when the operator is cluster-scoped).

The output for errors found during validation is also easier to read. 

Create a playbook to run the `aap_ocp_install` role with appropriate variables set (see the README.md in the role).

No

None

### DIFF
--- a/roles/aap_ocp_install/README.md
+++ b/roles/aap_ocp_install/README.md
@@ -11,31 +11,31 @@ In addition the kubernetes.core and redhat.openshift Ansible collections are req
 
 A description of the settable variables for this role should go here, including any variables that are in defaults/main.yml, vars/main.yml, and any variables that can/should be set via parameters to the role. Any variables that are read from other roles and/or the global scope (ie. hostvars, group vars, etc.) should be mentioned here as well.
 
-| Variable Name              | Default Value | Required | Description                                                            |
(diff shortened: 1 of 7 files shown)


sean-m-sullivan (approved):
I think last step is allowing them to be separately set, but this is great! works really well.

branic:
> I think last step is allowing them to be separately set

@sean-m-sullivan can you expand on this a little? I'm not quite sure what you are meaning.

sean-m-sullivan:
it may be I am alone, but was thinking this should be valid and vice versa
```
- name: Install AAP on OCP playbook
  hosts: localhost
  gather_facts: false

  vars:
    aap_ocp_install_ocp_connection:
      host: "https://api.crc.testing:6443"
      username: kubeadmin
      password: nYF8X-UBSDn-2hzzS-Tg52E
      validate_certs: false
    aap_ocp_install_namespace: aap-test
    aap_ocp_install_operator:
      channel: "stable-2.2"
    aap_ocp_install_controller:
      instance_name: automationcontroller

  roles:
    - redhat_cop.aap_utilities.aap_ocp_install
```

branic:
I agree, that should be a valid config (and I thought I tested a variation like that). I'll test again and fix what's not working.

PR #93, opened 2022-09-01
Allow individual components to be installed on OCP

branic:
### What does this PR do?

Allow each component to be individually installed (operator, controller, hub).

As an example the below playbook will now execute without generating an error for missing hub configuration info and successfully installs the operator and controller.

```
- name: Install AAP on OCP playbook
  hosts: localhost
  gather_facts: false

  vars:
    aap_ocp_install_ocp_connection:
      host: "https://api.crc.testing:6443"
      username: kubeadmin
      password: <REDACTED>
      validate_certs: false
    aap_ocp_install_namespace: aap-test
    aap_ocp_install_operator:
      channel: "stable-2.2"
    aap_ocp_install_controller:
      instance_name: automationcontroller

  roles:
    - redhat_cop.aap_utilities.aap_ocp_install
 ```

Automated tests are preferred, but not always doable - especially for infrastructure.
Include commands to run your new feature, and also post-run commands to validate that it worked. (please use code blocks to format code samples)

No

None

### DIFF
--- a/roles/aap_ocp_install/tasks/main.yml
+++ b/roles/aap_ocp_install/tasks/main.yml
@@ -23,6 +23,7 @@
     apply:
       tags:
         - operator
+  when: aap_ocp_install_operator is defined
   tags:
     - operator
 
@@ -32,6 +33,7 @@
     apply:
       tags:
         - controller
+  when: aap_ocp_install_controller is defined
   tags:
     - controller
 
@@ -41,6 +43,7 @@
     apply:
       tags:
         - hub
+  when: aap_ocp_install_hub is defined
   tags:
     - hub
 ...
--- a/roles/aap_ocp_install/README.md
+++ b/roles/aap_ocp_install/README.md
@@ -19,7 +19,8 @@ A description of the settable variables for this role should go here, including 
 | aap_ocp_install_controller | Yes*     | None          | Dictionary containing keys defined in the `controller variables table` |
 | aap_ocp_install_hub        | Yes*     | None          | Dictionary containing keys defined in the `hub variables table`        |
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by ericzolf** [comment]
I didn't go (yet) into the details of the review but I'm slightly worried by the explicit validation of `ansible_run_tags`.

Let me try to word my worries:

1. first time I see this, so always worrying (or a cool new thing to learn :+1: )
2. more seriously, checking `ansible_run_tags` without checking `ansible_skip_tags` could lead to unexpected behavior of tags, 
3. mixing tags and variables could lead to strange side effects
4. and then, overall, this could make the whole setup rather complex
5. and, in the good practices, we recommend _against_ tags within roles

I'm not fully against it, I'm just wondering if it's a good idea, and open to discussion.

**Comment by branic** [comment]
I can remove the tags and just base the task execution on the existence of the variables. The tags do add a level of complexity.

File: roles/aap_ocp_install/tasks/main.yml [hunk 1]
```diff
@@ -32,6 +34,8 @@
     apply:
       tags:
         - controller
+  when:
+    - ( 'controller' in ansible_run_tags ) or ( 'all' in ansible_run_tags and aap_ocp_install_controller is defined )
```
sean-m-sullivan: should be sufficient for the If defined, if someones running all it will do the right thing, if they are only running one tag or the other and the variable, it will do the right thing. No need to check the tags in the when. I think they take care of themselves.

branic: I simplified the when conditions to only look for the variable to be defined. :man_facepalming: 

I tested with various combinations of variables defined and specifying tags.

branic:
I finally got back around to working on this PR. I rebased it so there wouldn't be any merge conflicts as its been a while. I also simplified the when conditions and removed the use of `ansible_run_tags`.

sean-m-sullivan (approved):
LGTM

PR #95, opened 2022-09-10
Update linter versions

branic:
### What does this PR do?

Updates the pre-commit-hooks linter versions and fixes reported errors.

Automated tests are preferred, but not always doable - especially for infrastructure.
Include commands to run your new feature, and also post-run commands to validate that it worked. (please use code blocks to format code samples)

No

None


sean-m-sullivan (approved):
Thanks for doing this, we need to stay on top of our linting!

PR #98, opened 2022-09-16
enforce the version as string so user doesn't need to

djdanielsson:
### What does this PR do?
should remove the version being a string requirement

test with version being a string or number and make sure the collection still functions as expected

no

none


sean-m-sullivan (approved):
LGTM

PR #104, opened 2022-10-14
Fix issues with copy and /dev/null

ericzolf:
### What does this PR do?

ansible.builtin.copy doesn't properly handle /dev/null as source, it removes it if called as root. See https://github.com/ansible/ansible/issues/79101 for details

Run or better re-run aap_setup_install

n/a

https://github.com/ansible/ansible/issues/79101


djdanielsson (approved):
LGTM

PR #110, opened 2022-11-29
remove warn

sean-m-sullivan:
# What does this PR do?
Removes warn as AAP 2.3 requires ansible 2.14, and it removed this module option.


djdanielsson (approved):
LGTM

PR #123, opened 2023-01-18
Fix tag

### LINKED ISSUE #97: Incorrect tag???
https://github.com/redhat-cop/aap_utilities/blob/81be4b98694fb527db51ef1c94511f2131d932e7/roles/aap_remove/tasks/main.yml#L13

I think this tag should not be `controller`, it should probably be something like `hub`.

As a side-note based on a comment from @ericzolf (https://github.com/redhat-cop/aap_utilities/pull/93#issuecomment-1242717206) maybe we should instead look at removing all tags from the roles. Thoughts?

Tompage1994:
# What does this PR do?

Fixes the tag within the aap_remove role
# How should this be tested?
Can use `hub` or `ah` as a tag to only run that section.

resolves #97

N/A

### DIFF
--- a/changelogs/fragments/fix_tag.yml
+++ b/changelogs/fragments/fix_tag.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Fixed tag in aap_remove role for automation hub.
+...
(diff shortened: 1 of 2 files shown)


sean-m-sullivan (approved):
good catch

PR #124, opened 2023-01-18
remove references to Tower and change occurances of redhat_cop to infra

### LINKED ISSUE #99: Backup role does not add much value
Looking at the [backup](./roles/backup/) role, it is just running the backup command (setup.sh -b). Is there something else the role can perform? otherwise I could just run the setup.sh command myself.  Also, I'm not sure why the role asks for Postgres credentials etc when all of this info should be in the `inventory` file and leveraged using `setup.sh -i inventory -b` command. 

Also, of course, need to rewrite the role to talk about AAP instead of Tower.

Tompage1994:
# What does this PR do?

I started this removing any old Tower references but then also made changes for redhat_cop -> infra namespace

# How should this be tested?

Readme changes which CI will lint

resolves #99 
Though only some of the additional comments. The original aim of the issue probably isn't required to be addressed.

N/A


sean-m-sullivan (approved):
LGTM nice work!

PR #126, opened 2023-01-19
Add meta-unsupported-ansible to skip-list of ansible lint

Tompage1994:
# What does this PR do?

Fixes broken CI Ansible-lint where meta-unsupported-ansible is presumably a new rule which we don't match but we don't need to change out minimum supported Ansible so we should just skip this rule.

# How should this be tested?
CI will pass

N/A

See failing CI here: https://github.com/redhat-cop/aap_utilities/actions/runs/3955727871/jobs/6774258851


branic (approved):
LGTM

PR #133, opened 2023-02-15
Fix hostnames for install check and bump version to 2.3

ericzolf:
# What does this PR do?

The hostname defaults for controller and PAH were assuming a list were a dictionary was expected. Took the chance to bump the default AAP version to 2.3

# How should this be tested?

Try to update/install a cluster with controller and AH, and without forcing the installation. Without this fix, it fails stating that the dictionary of hosts doesn't have a `[0]` index.

n/a - found out while updating my own cluster

The fix isn't necessary anymore according to https://issues.redhat.com/browse/AAP-4555

### DIFF
--- a/roles/aap_setup_download/defaults/main.yml
+++ b/roles/aap_setup_download/defaults/main.yml
@@ -7,7 +7,7 @@
 # aap_setup_down_offline_token:
 
 # which version of the installer to download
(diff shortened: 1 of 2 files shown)


Tompage1994 (approved):
LGTM

PR #136, opened 2023-02-16
Skip ansible-lint rules

Tompage1994:
# What does this PR do?
Fixes lint issues caused by update to ansible-lint 6.13.0

# How should this be tested?
CI should pass

N/A

Note that I've set jinja[invalid] to warn for now because the new version has caused a regression in this rule which I've raised a bug for and linked to from the PR in controller_configuration https://github.com/redhat-cop/controller_configuration/pull/503

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -4,6 +4,9 @@
 # https://github.com/ansible/ansible-lint/issues/371
 # exclude_paths:
 #   - roles/master_role_example/
+exclude_paths:
+  - '.github/'
+  - 'changelogs/'
 parseable: true
 use_default_rules: true
 # https://github.com/ansible/ansible-lint/issues/808
@@ -11,4 +14,7 @@ use_default_rules: true
 # verbosity: 1
 skip_list:
   - meta-runtime
+warn_list:
+  - jinja[invalid]  # Temporarily adding this due to https://github.com/ansible/ansible-lint/issues/3048
+  - jinja[spacing]  # Would be set to warn by default and the only flagged issues are fairly unfixable
 ...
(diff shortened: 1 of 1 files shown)


.ansible-lint
```diff
@@ -4,11 +4,17 @@
...
 parseable: true
 use_default_rules: true
 # https://github.com/ansible/ansible-lint/issues/808
 #  with verbosity set to 1, its dumping 'unknown file type messages'
 # verbosity: 1
 skip_list:
   - meta-runtime
+warn_list:
+  - jinja[invalid]  # Temporarily adding this due to https://github.com/ansible/ansible-lint/issues/3048
+  - jinja[spacing]  # Would be set to warn by default and the only flagged issues are fairly unfixable
```
djdanielsson: Why are we not failing on spacing here?

Tompage1994: It started failing when I added this warn_list, so I assume it was set as a warn rule before. It's only a couple of lines and if you look at them they can't really be fixed anyway

Tompage1994: I also realise that may not have been on this repo but it made sense to make all the repos skip the same rules

sean-m-sullivan (approved):
For now, we should look into this more, but the jinja spacing as Tom has said likely not fixable.

sean-m-sullivan:
@Tompage1994 Needs a rebase

branic:
If there are places where `jinja[spacing]` can't be fixed then we should turn off the check just for that case and not for the entire project.

Tompage1994:
> If there are places where `jinja[spacing]` can't be fixed then we should turn off the check just for that case and not for the entire project.

You're right. I've raised PRs or changed PRs in other repos to fix this. I've updated this PR too

djdanielsson (approved):
LGTM

branic (approved):
LGTM

PR #150, opened 2023-03-30
handle passwords in aap_setup_prep_inv_secrets

jce-redhat:
# What does this PR do?

modifies the aap_setup_install role to use credentials from the aap_setup_prep_inv_secrets variable if defined

Brief explanation of the code or documentation change you've made

the aap_setup_install role uses controller and hub credentials when checking that the components are running both before and after the setup script is run.  however, it only used credentials defined in `aap_setup_prep_inv_vars`, but according to the aap_setup_prep role they can also be defined in `aap_setup_prep_inv_secrets`.  this PR modifies the tasks to use either, preferring credentials in `aap_setup_prep_inv_secrets` over those in `aap_setup_prep_inv_vars`.

# How should this be tested?

only set controller and hub credentials in `aap_setup_prep_inv_secrets` when installing AAP.  before this change, running the aap_setup_install role would fail when checking if controller or hub were running if the credentials were only set in this variable.

no

resolves #[number]

n/a

n/a

### DIFF
--- a/roles/aap_setup_install/tasks/main.yml
+++ b/roles/aap_setup_install/tasks/main.yml
@@ -11,8 +11,7 @@
     url: "https://{{ controller_hostname }}/"
     method: GET
     user: admin
-    password: "{{ aap_setup_prep_inv_secrets.all.admin_password | default(false) or
-                  aap_setup_prep_inv_vars.all.admin_password }}"
+    password: "{{ aap_setup_prep_inv_secrets.all.admin_password | default(aap_setup_prep_inv_vars.all.admin_password) }}"
     validate_certs: "{{ controller_validate_certs | default('false') }}"
     force_basic_auth: true
   register: __aap_setup_inst_ctl_check
@@ -27,8 +26,7 @@
     url: "https://{{ ah_hostname }}/api/galaxy/"
     method: GET
     user: admin
-    password: "{{ aap_setup_prep_inv_secrets.all.automationhub_admin_password | default(false) or
-                  aap_setup_prep_inv_vars.all.automationhub_admin_password }}"
+    password: "{{ aap_setup_prep_inv_secrets.all.automationhub_admin_password | default(aap_setup_prep_inv_vars.all.automationhub_admin_password) }}"
     validate_certs: "{{ controller_validate_certs | default('false') }}"
     force_basic_auth: true
   register: __aap_setup_inst_ah_check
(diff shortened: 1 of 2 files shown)


roles/aap_setup_install/tasks/main.yml
```diff
@@ -11,7 +11,8 @@
     url: "https://{{ controller_hostname }}/"
     method: GET
     user: admin
-    password: "{{ aap_setup_prep_inv_vars.all.admin_password }}"
+    password: "{{ aap_setup_prep_inv_secrets.all.admin_password | default(false) or
```
Tompage1994: Is this right? I assume you want to fallback to `aap_setup_prep_inv_vars.all.admin_password` but my reading of this is that it will just default(false) which is presumably not what we want to do here?

Is the correct solution here not:

```
    password: "{{ aap_setup_prep_inv_secrets.all.admin_password | default(aap_setup_prep_inv_vars.all.admin_password) }}"
```

roles/aap_setup_install/tasks/main.yml
```diff
@@ -26,7 +27,8 @@
     url: "https://{{ ah_hostname }}/api/galaxy/"
     method: GET
     user: admin
-    password: "{{ aap_setup_prep_inv_vars.all.automationhub_admin_password }}"
+    password: "{{ aap_setup_prep_inv_secrets.all.automationhub_admin_password | default(false) or
```
Tompage1994: Similar question here

Tompage1994 (changes requested):
Some questions about what you're doing.

It would also be good if the README was updated to reflect this change in variable naming.

Also a changelog will need to be added for this

roles/aap_setup_install/tasks/main.yml
jce-redhat: yup, that was an artifact of me trying something else, defaulting to `aap_setup_prep_inv_vars.all.admin_password` is the way.

roles/aap_setup_install/tasks/main.yml
jce-redhat: similar reply.  :)

jce-redhat:
thanks for the feedback @Tompage1994!  i wasn't sure about the changelog snippet since i didn't know if you roll up PRs into releases or if you want something per change.
EDIT: but now that i think about it, i guess the fragments need to be there for when a release is cut....

the variables are already documented in (roles/aap_setup_prep/README.md), but i found if i set the controller and hub passwords in `aap_setup_prep_inv_secrets` only, they were never referenced during the checks for controller and hub availability.  that's what lead to this PR.

Tompage1994 (approved):
LGTM, thanks

PR #171, opened 2023-05-19
Changed example, added comment, noting URL truly is required

jeffwarnica:
Updates two sample variable files to have what are required to be URLs in URL format.


djdanielsson (approved):
I will just push this through

PR #174, opened 2023-06-07
update lint

sean-m-sullivan:
# What does this PR do?

Fix linting
we use unique variables and __ as a prefix, but disabling this check for now


djdanielsson (approved):
LGTM

PR #182, opened 2023-07-20
Added EDA to OCP installation role.

zaskan:
# What does this PR do?

Adds EDA option to OCP installation


sean-m-sullivan (approved):
This looks good to me, @branic I don't have a OCP to test on atm, can you give this a once over?

roles/aap_ocp_install/templates/eda/instance.yaml.j2
```diff
@@ -0,0 +1,21 @@
...
+spec:
+  ui:
+    replicas: 1
+  set_self_labels: true
+  api:
+    gunicorn_workers: 2
+    replicas: 1
+  redis:
+    replicas: 1
+  automation_server_url: "https://{{ __aap_ocp_install_controller_route }}"
```
branic, line 16: It would be good to have the `automation_server_url` also be able to be a user supplied value for cases when the controller has already been installed and EDA is just being added. If the controller isn't being installed (e..g. the `eda` tag is used then the `__aap_ocp_install_controller_route` won't be defined.

Also, need to determine what happens if the variable is not set and the controller was not just installed as part of the same run. We don't want it to blow up because of an undefined variable.

roles/aap_ocp_install/tasks/pre-validate-eda.yml
branic, line 1: The inclusion of this file needs to be added to `pre-validate.yml`. Follow how lines 69-73 are used to include the automation hub validation steps.

branic (changes requested):
@zaskan Thanks for this PR and sorry it has taken me so long to review. There are a few changes needed before this PR can be merged.

In addition to the code comments please also:

- update the `README.md` for the role to document the EDA variables and any other pertinent information.
- Add a changelog fragment in `/changelogs/fragments/<FILENAME>.yml` to note the addition.

djdanielsson:
@zaskan thanks for the PR, if you could update it with the requested changes we can get this change merged

djdanielsson:
if you could do the requested changes so we can merge this or I will close the PR

djdanielsson:
closing due to changes never happening and its been almost a year

PR #183, opened 2023-08-02
Make galaxy.yml persistent and move out ansible.cfg

Tompage1994:
# What does this PR do?
Makes galaxy.yml persistent and moves the template out of root. 

This makes it installable from source

# How should this be tested?
1. Test we can install from source
2. We will need to perform a release and ensure that the galaxy.yml remains in the repo with `-devel` appended to the version that's just been released. We need to also be sure that the galaxy.yml of the released version is correct

related https://github.com/ansible/automation_hub_collection/issues/280


djdanielsson (approved):
LGTM

PR #192, opened 2023-09-05
fix broken example

jon4hz:
This PR fixes a broken example in the `aap_setup_install` README.

### DIFF
--- a/roles/aap_setup_install/README.md
+++ b/roles/aap_setup_install/README.md
@@ -55,18 +55,18 @@ aap_setup_rhel_version: 8
 
 aap_setup_prep_inv_nodes:
   automationcontroller:
-    - ansible-ctrl.example.com
+    ansible-ctrl.example.com:
   automationhub:
-    - ansible-hub.example.com
+    ansible-hub.example.com:
   automationedacontroller:
-    - ansible-eda.example.com
+    ansible-eda.example.com:
   database:
-    - database.example # if using an already existing DB, remove this and ensure that the following variables are filled with the valid details for your Controller and PAH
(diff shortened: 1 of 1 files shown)


jon4hz:
Also I noticed that the `ansible_become` [here](https://github.com/redhat-cop/aap_utilities/blob/10bd400f1e3e86376aaaf4c41d6f5057f8e8599a/roles/aap_setup_install/README.md?plain=1#L83) overrides  the `become: false` in the role `automation_platform_installer.config_dynamic` L58 which can lead to some issues. For example if you use an ee to execute that role, the task will most likely fail due to the missing sudo package inside the ee.

sean-m-sullivan:
Not sure what we are trying to fix here, these are set to be list items, not dict items, [and the template is working around lists, ](https://github.com/redhat-cop/aap_utilities/blob/bdc76b0071c5f54cfd32424c1ba22503492c1539/roles/aap_setup_prepare/templates/inventory.j2#L6)

could you post what happened that led to an error for this fix?

This has worked for me in the past
```
aap_setup_prep_inv_nodes:
  database:
    - 'hub.laptop'
  automationhub:
    - 'hub.laptop'
```

jon4hz:
Actually, I'm not sure about that anymore either.  When I started working with the `aap_setup_prepare` role, I got a weird template error when the inv_nodes were a list, but I guess the error was caused by something else, although I was pretty sure that changing the list to dict items solved the problem. 

I tested my setup again with the nodes provided as a list and now it works, weird. Anyway... sorry for the confusion.

sean-m-sullivan:
No worries, my worry was more if you had more then one node, the dict option would likely not work, but if you find that error again let us know

Issue #206, opened 2023-10-20
aap_setup_download role fails to download setup bundle


snagoor:
# Summary

infra.aap_utilities.aap_setup_download role fails to find and download the correct bundle as things change on the Red Hat CDN

# Issue Type

- Bug Report

# Ansible, Collection, Docker/Podman details

infra.aap_utilities   2.3.0 version 

```console (paste below)
ansible --version
ansible [core 2.15.5]
  config file = /var/lib/awx/projects/awxlab/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /var/lib/awx/projects/awxlab/collections
  executable location = /usr/bin/ansible
  python version = 3.9.16 (main, Sep 12 2023, 00:00:00) [GCC 11.3.1 20221121 (Red Hat 11.3.1-4)] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True

ansible-galaxy collection list
Collection          Version
------------------- -------
ansible.posix       1.5.4  
awx.awx             23.0.0 
community.crypto    2.15.1 
community.general   7.3.0  
infra.aap_utilities 2.3.0  
kubernetes.core     2.4.0  
nshaik.aap          1.0.0 

podman --version
podman version 4.4.1
```

- ansible installation method: one of source, pip, OS package, EE
- RPM Package

# OS / ENVIRONMENT

# Desired Behavior

# Actual Behavior

Please give some details of what is actually happening.
Include a [minimum complete verifiable example] with:

- playbook / task
- configuration file / list
- error

```console (error)
TASK [infra.aap_utilities.aap_setup_download : Downloading the latest installer of type setup] ***
skipping: [aap.example.com] => (item={'imageName': 'Ansible Automation Platform 2.3 Setup Bundle', 'filename': 'ansible-automation-platform-setup-bundle-2.3-2.9.tar.gz', 'arch': 'x86_64', 'datePublished': '2023-10-17T16:55:44.000Z', 'checksum': 'ea1e70775533cedac9a9d331d7446b0dc2f555820187a7b15ed398e3da5bb00d', 'downloadHref': 'https://xxxxx/download'}) 
[...]

Issue #199, opened 2023-09-26
Incorrect installation setup bundle extracted and configured


snagoor:
# Summary

`infra.aap_utilities.aap_setup_download` downloads 2 different setup types 

1. ansible-automation-platform-containerized-setup-2.4-1.tar.gz
2. ansible-automation-platform-setup-2.4-2.tar.gz

# Issue Type

- Bug Report

# Ansible, Collection, Docker/Podman details

```console (paste below)
ansible --version
ansible [core 2.14.2]
  config file = /root/aap/aap_standalone/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.11/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.11.2 (main, May 24 2023, 00:00:00) [GCC 11.3.1 20221121 (Red Hat 11.3.1-4)] (/usr/bin/python3.11)
  jinja version = 3.1.2
  libyaml = True

ansible-galaxy collection list

# /root/.ansible/collections/ansible_collections
Collection            Version
--------------------- -------
ansible.posix         1.5.4  
awx.awx               23.0.0 
community.crypto      2.15.1 
community.general     7.3.0  
infra.aap_utilities   2.3.0  
kubernetes.core       2.4.0  
nshaik.aap_collection 1.0.14 

podman --version
podman version 4.4.1

```

- ansible installation method: one of source, pip, OS package, EE

# OS / ENVIRONMENT

Ansible Controller installation on RHEL 9.2 version

# Desired Behavior

Installation inventory file should be configured in the correct installation bundle extracted folder

# Actual Behavior

Installation bundle extracted and inventory file is configured in wrong location `/var/tmp/ansible-automation-platform-containerized-setup-2.4-1/` where the correct collection  should be `/var/tmp/ansible-automation-platform-setup-2.4-2/`

```console (error)

```

# STEPS TO REPRODUCE

Sample variable file that was used to call `infra.aap_utilities.aap_setup_download` role
```
aap_setup_down_offline_token: "xxxxeyJhbGciOixxxx"
aap_setup_down_version: "2.4"
aap_setup_down_type: "setup"
aap_setup_rhel_version: "9"
aap_setup_prep_inv_nodes:
  automationcontroller:
    - controller.example.com
[...]

PR #207, opened 2023-10-20
fixes problems with AAP setup downloads

snagoor:
This PR solves both issues report in #206 and #199


djdanielsson (approved):
I think this looks good, I haven't had time to test it. This won't work with the container option but I think for now we want that anyways.

PR #215, opened 2024-01-15
Remove SSO and add EDA variables

ericzolf:
# What does this PR do?

Mainly do the necessary changes in the defaults/main.yml of aap_setup_prepare, because EDA is (relatively) new and SSO is outdated.

# How should this be tested?

Nothing to be tested, just documentation and defaults which are commented out.

n/a

n/a


sean-m-sullivan (approved):
Thanks for this Eric, Needed to be done!

PR #225, opened 2024-03-25
Support the containerized AAP setup/setup-bundle

spyrexd:
Install and configure AAP using the containerized installer.

- The installer arch is now configurable.
- aap_remove not updated because the containerized installer has an uninstall playbook.

# What does this PR do?

Enables installing APP using the containerized installer

# How should this be tested?

Used the following configuration to test deploying AAP on  a RHEL 9 VM running on MacOS 

```yaml
---
aap_setup_down_type: "setup"
aap_setup_rhel_version: 9
aap_setup_arch: aarch64
aap_setup_down_offline_token: UPDATE
aap_setup_containerized: true
controller_validate_certs: false
ah_validate_certs: false
eda_validate_certs: false


aap_setup_prep_inv_nodes:
  automationcontroller:
    aap-test.example.com:
  automationhub:
    aap-test.example.com:
  automationeda:
    aap-test.example.com:
  database:
    aap-test.example.com:


aap_setup_prep_inv_vars:
  automationcontroller:
    ansible_connection: local
  automationhub:
    ansible_connection: local
  automationeda:
    ansible_connection: local

  all:
    controller_admin_password: admin
    controller_pg_host: 127.0.0.1
    controller_pg_password: admin
    hub_admin_password: admin
    hub_pg_host: 127.0.0.1
    hub_pg_password: admin
    postgresql_admin_password: admin
    eda_admin_password: admin  
    eda_pg_host: 127.0.0.1
    eda_pg_password: admin
    registry_username: UPDATE
    registry_password: UPDATE

```

resolves #224 


roles/aap_setup_install/defaults/main.yml
```diff
@@ -66,7 +69,7 @@ ah_hostname: "{{ (aap_setup_prep_inv_nodes['automationhub'].keys() | list)[0] }}
 # ah_oauthtoken: ""
 # ah_validate_certs: false
 
-eda_hostname: "{{ (aap_setup_prep_inv_nodes['automationedacontroller'].keys() | list)[0] }}"
+eda_hostname: "{{ (aap_setup_prep_inv_nodes[__aap_setup_inst_eda_group_name].keys() | list)[0] }}"
```
djdanielsson, line 72: this is the one part just looking over quickly that seems wrong and I would like to double check because the other components are not like this line.

roles/aap_setup_install/vars/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+# Set the https port used for validating services are up
+__aap_setup_inst_controller_port: "{{ aap_setup_prep_inv_vars.automationcontroller.nginx_https_port | default('8443') }}"
```
djdanielsson, line 3: is there a reason why this is in vars instead of defaults like all the other vars?

djdanielsson (changes requested):
looks pretty good, I did have 2 questions at least from my quick skim over it.

roles/aap_setup_install/defaults/main.yml
spyrexd: The group name for eda is different between the regular setup installer and the containerized installer. It is one of the inconstancies between the installer inventory.yml files.

roles/aap_setup_install/vars/main.yml
spyrexd: These values are not expected to be overridden/passed in by the playbook calling the setup install role.  The ports are only changed if the `nginx_https_port` is set in the `aap_setup_prep_inv_vars`. Other wise the default port numbers are used.

spyrexd: This brings up another question. Currently the default ports are the TLS enabled ports. Should these point to the http ports and allow the redirect point to the correct TLS enabled port?

roles/aap_setup_install/defaults/main.yml
djdanielsson: but then wouldn't this line be different for all of them for container option but not for setup? and if so then wouldn't this be wrong for setup?

roles/aap_setup_install/vars/main.yml
djdanielsson: but why wouldn't they be expected to be over written? I think a lot of people might not want to use the default ports the container installer uses. This is also related to your TLS enabled port question.

roles/aap_setup_install/defaults/main.yml
spyrexd: `__aap_setup_inst_eda_group_name` is set in the role vars: 

```yaml
_aap_setup_inst_eda_group_name: "{{ (aap_setup_inst_containerized | bool) | ternary('automationeda', 'automationedacontroller') }}"
```

If we are not a containerized install then it is set to 'automationedacontroller'

roles/aap_setup_install/vars/main.yml
spyrexd: To set the ports nginx uses during the install, the vars `nginx_http_port` and `nginx_https_port` need to be set in the `inventory.yml`.  We can set these vars in the `inventory.yml` during the prep step by adding the vars to a host group when configuring `aap_setup_prep_inv_vars`. 

One key difference with the containerized installer is that you can run all the services on a single host. By default all the services are configured to listen on non-standard ports.  Because of this, we need to be able to determine if `nginx_http_port` or `nginx_https_port` are set. If they are set then we need to use the values when validating the servies are running or use the default ports below.

| App | HTTP | HTTPS |
|-----|-------|--------|
| Controller | 8080 | 8443 |
| Hub | 8081 | 8444 |
| EDA | 8082 | 8445 |

There are additional complications when installing on a single node. You can not set `nginx_http_port` or  `nginx_https_port` as it will assign the same port for all services. This can be worked around with multiple host entires in the /etc/hosts for the same host. Also the default ports changed between the current version of the installer and previous versions.

roles/aap_setup_install/defaults/main.yml
djdanielsson: Ok but it doesn't look like it's set that way for hub and controller shouldn't it be the same?

roles/aap_setup_install/vars/main.yml
djdanielsson: Yes I know which is why I think we should make this be an expected variable from the user to set what port they want. I will double check my container inventory because I haven't messed with it much since I configured it in the beta release.

djdanielsson: I do understand what you mean about the nginx_http(s)_port var and if it's the same host it will mess stuff up. I just wonder if we should make a check that verifies that if they set port it cannot have the same hosts... I would be interested in hearing what others think about how we should solve/set this

djdanielsson: @sean-m-sullivan @Tompage1994 @ericzolf

roles/aap_setup_install/defaults/main.yml
djdanielsson: ah nvm I understand now what you mean

roles/aap_setup_install/vars/main.yml
sean-m-sullivan: I think as is, is fine, we could add the feature in the future if we add the other logic around setting multiple hosts, but I believe the containerized is supposed to be single node by design.

sean-m-sullivan (approved):

djdanielsson (approved):

PR #238, opened 2024-06-13
Make inventory file mode 600

### LINKED ISSUE #237: inventory file is world-readable - Can we make it only user-readable?
Hi.

The inventory file that aap_setup_prepare generates contains a lot of credentials, but is world-readable.

Wouldn't it be better from a security standpoint to make it only readable to the user it belongs?

Shouldn't be problematic to change that in terms of running the setup.sh script.

Yours
  David

dbk-rabel:
The inventory file that aap_setup_prepare generates contains a lot of credentials, but has very unsafe permissions (664) at the moment.

This commit changes it to 600 permissions.

fixes #237

### DIFF
--- a/roles/aap_setup_prepare/tasks/main.yml
+++ b/roles/aap_setup_prepare/tasks/main.yml
@@ -34,7 +34,7 @@
   ansible.builtin.template:
     src: inventory.j2
     dest: "{{ aap_setup_prep_setup_dir }}/inventory"
-    mode: "ug=rw,o=r"
+    mode: "600"
   when: aap_setup_prep_process_template | bool
 
 - name: Apply relevant fixes / workarounds
--- a/roles/aap_setup_prepare/tasks/main.yml
+++ b/roles/aap_setup_prepare/tasks/main.yml
@@ -34,7 +34,7 @@
   ansible.builtin.template:
     src: inventory.j2
     dest: "{{ aap_setup_prep_setup_dir }}/inventory"
-    mode: "600"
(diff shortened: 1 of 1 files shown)


djdanielsson:
I am trying to think of any scenario where this could break something

dbk-rabel:
> I am trying to think of any scenario where this could break something

Well, theoretically someone could try to start setup.sh with a different user. So in theory it could break something.

But I guess thats very uncommon. If people use aap_setup_prepare, they probably also use aap_setup_install and will therefore probably use the same ansible user.

djdanielsson:
yea I had a similar thought... how about this change it to 640 and we go with that? that way if someone wants to use a different user to deploy or something they could just add them to the same group and it should just work

dbk-rabel:
> yea I had a similar thought... how about this change it to 640 and we go with that? that way if someone wants to use a different user to deploy or something they could just add them to the same group and it should just work

Yes, that should also be fine. :)

roles/aap_setup_prepare/tasks/main.yml
```diff
@@ -34,7 +34,7 @@
   ansible.builtin.template:
     src: inventory.j2
     dest: "{{ aap_setup_prep_setup_dir }}/inventory"
-    mode: "ug=rw,o=r"
+    mode: "600"
```
dbk-rabel:
```suggestion
    mode: "640"
```

djdanielsson (approved):
LGTM

PR #240, opened 2024-06-18
fix download, add containerized download documentation

### LINKED ISSUE #226: Add support for AAP containerized installer
### Discussed in https://github.com/redhat-cop/aap_utilities/discussions/224

<div type='discussions-op-text'>

<sup>Originally posted by **spyrexd** March 25, 2024</sup>
Support using the containerized installer when downloading and installing AAP. 

Allow setting the arch of the installer to enable containerized installs to VMs running on aarch64.</div>

sean-m-sullivan:
# What does this PR do?

bugfixes:
  - Fixed an issue where where download would error trying to download a second version of the installer.
 
minor_changes:
  - Added ability to download the containerized installer.
  - Tweaked how the installer to download is choosen, this should have no effective changes.

# How should this be tested?

Already tested, this will error out if someone tries to download the containerized for rhel 8, as it doesn't exist, but don't have a great way of preventing that.

resolves #226

### DIFF
--- a/changelogs/fragments/download_fix_and_update.yml
+++ b/changelogs/fragments/download_fix_and_update.yml
@@ -1,0 +1,7 @@
+---
+bugfixes:
+  - Fixed an issue where where download would error trying to download a second version of the installer.
+minor_changes:
+  - Added ability to download the containerized installer.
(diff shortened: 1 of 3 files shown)


djdanielsson (approved):
LGTM

PR #251, opened 2024-07-28
fix generated inventory files without a database inventory group

AdilMetkali:
Without the control on the existence of the groups in the inventory file, not having a database section in the inventory generates a malformed inventory file.

# What does this PR do?

Without the control on the existence of the groups in the inventory file, not having a database section in the inventory generates a malformed inventory file.


djdanielsson (approved):
LGTM

PR #261, opened 2024-11-21
Add support for new platform installation method of AAP 2.5 on OpenShift

### LINKED ISSUE #260: Update the infra.aap_utilities.aap_ocp_install module to support AAP 2.5
The current aap_ocp_install role support installation of AAP versions up to 2.4. In the AAP 2.5 release the method of deployment on OpenShift has changed to use a new "platform" CR which installs all of the AAP components and the platform gateway.

We should support installation of AAP on OCP using either the old (2.4 and earlier) method, or the new (2.5 and later) platform installation method using the aap_ocp_install role.

Details of the 2.5 operator installation method can be found here:

https://docs.redhat.com/en/documentation/red_hat_ansible_automation_platform/2.5/html/installing_on_openshift_container_platform/configure-aap-operator_operator-platform-doc#operator-link-components_configure-aap-operator

derekwaters:
# What does this PR do?

This PR modifies the aap_ocp_install role to support the new platform installation method introduced to the AAP operator in AAP 2.5. It is backward 

# How should this be tested?

Requires an OpenShift cluster with appropriate credentials, then run:

```
      ansible.builtin.include_role:
        name: "infra.aap_utilities.aap_ocp_install"
```

with configuration as specified in the README and vars files defined in the role. 

Once complete, installation can be validated by accessing AAP via the quick links toolbar menu option.

https://github.com/redhat-cop/aap_utilities/issues/260
resolves #260 

None


galaxy.yml
```diff
@@ -1,7 +1,7 @@
 ---
 namespace: infra
 name: aap_utilities
-version: 2.5.2-devel
+version: 1.0.0
```
djdanielsson: this is not correct

djdanielsson (dismissed):
there is at least one change that needs to happen, I don't have OCP to test this so I will wait for someone who does to review and approve or not

galaxy.yml
derekwaters: @djdanielsson I thought so too, but the pre-commit tests seemed to be flagging that as an issue? I'll change it back.

djdanielsson: It might not like the devel but we want to keep that because of how we have stuff working if pre-commit doesn't like it then ignore it

branic:
I'm planning on reviewing and testing, I'll should be able to get to it in the next day or two.

branic (changes requested):
@derekwaters Thank you for your patience with me reviewing this PR.

I'd liek to request a couple of changes. I've been to get a working installation, but so far have not been successful. The Operator is installed and working and I've been able to get a platform manifest created and it started with a controller defined, but for several minutes after the playbook completes the web ui fails during login (I think because the controller instance is not ready yet). So I think there needs to be some other checks to ensure that the system is ready to be used before the playbook ends.

Also, it would be good to support installing into a different namespace than the operator is installed in.

I'm still getting my head around how the 2.5 install works and will provide more feedback.

One last request is to add an example playbook for installing the 2.5 version. Maybe relabel the existing example to be 2.4 and below and the new example as 2.5 and above.

derekwaters:
This PR has been updated to include:

1) The ability to install the operator and the platform in separate namespaces (when cluster-scoped operator channel is used)
2) Additional checks for controller / eda / hub API availability before proceeding with the playbook
3) Updated README to include separate example playbooks for 2.4 and earlier, and 2.5 and later

djdanielsson:
@branic have you had time to test it?

roles/aap_ocp_install/README.md
```diff
@@ -78,6 +83,13 @@ If the variable is omitted the corresponding component will not be installed (e.
 | link_text                          |          | Automation Hub (<INSTANCE_NAME>) | Text used for creating the OCP application link                   |
 | hub_manifest_overrides             |          | None                             | YAML Manifest to override the generated `AutomationHub` resource  |
 | consolelink_manifest_overrides     |          | None                             | YAML Manifest to override the generated `ConsoleLink` resource    |
+| storage_type                       |          | file                             | Hub storage type (file, S3 or azure)*                             |
```
branic: The default value isn't being taken into account. For example if specifying an instance name like:
```
    aap_ocp_install_hub:
      instance_name: tst-hub
```

The an error is encountered that the storage_type must be set:
```
TASK [infra.aap_utilities.aap_ocp_install : Ensure no validation errors found] ****************************************************************************
fatal: [localhost]: FAILED! => {
    "msg": [
        "The following errors must be fixed to continue:",
        [
            "aap_ocp_install_hub['storage_type'] must be 'file', 'S3' or 'azure'"
        ]
    ]
}
```

roles/aap_ocp_install/README.md
```diff
@@ -78,6 +83,13 @@ If the variable is omitted the corresponding component will not be installed (e.
...
+| file_storage_storage_class         |          | None                             | OpenShift StorageClass to use for file storage type for hub*      |
+| file_storage_size                  |          | 10Gi                             | Storage size for file storage type for hub*                       |
+| object_storage_s3_secret           |          | None                             | Name of an OpenShift Secret used to access S3 storage for hub*    |
+| object_storage_azure_secret        |          | None                             | Name of an OpenShift Secret used to access Azure storage for hub* |
```
branic: I think it would be clearer (and match other parts of the readme) if the `*` was moved to the `Required` column instead of being at the end of the description.

roles/aap_ocp_install/README.md
```diff
@@ -40,12 +43,14 @@ If the variable is omitted the corresponding component will not be installed (e.
 
 | Key Name                         | Required  | Default Value | Description                                                         |
 |----------------------------------|:---------:|---------------|---------------------------------------------------------------------|
-| channel                          | Yes       | None          | Channel to subscribe (e.g. stable-2.2 or stable-2.2-cluster-scoped) |
+| channel                          | Yes       | None          | Channel to subscribe (e.g. stable-2.2 or stable-2.2-cluster-scoped)*|
```
branic: I think it would be clearer (and match other parts of the readme) if the `*` was moved to the `Required` column instead of being at the end of the description.

NOTE: This applies to other sections where the `*` is in the description column.

roles/aap_ocp_install/README.md
```diff
@@ -90,13 +102,25 @@ If the variable is omitted the corresponding component will not be installed (e.
...
+### aap_ocp_install_platform keys
+
+| Key Name      | Required | Default Value                    | Description                                     |
+|---------------|:--------:|----------------------------------|-------------------------------------------------|
+| instance_name | Yes      | None                             | Name of the AAP Platform instance to create     |
+
+> ℹ️ **NOTE**
+>
+> These settings are only used when installing AAP 2.5 or later. instance_name and link_text values for individual components (hub, controller, eda) will be ignored when using the platform installation method.
+
```
branic: A similar `*` note should also be made for the `instance_name` and `link_text` name in the tables for each component key.

Also, this note and notes in other sections needs to include `namespace` as well as that is ignored in a platform install and all installed components are installed in the platform namespace.

roles/aap_ocp_install/README.md
```diff
@@ -90,13 +102,25 @@ If the variable is omitted the corresponding component will not be installed (e.
...
 ## Dependencies
 
 This role depends on the redhat.openshift and kubernetes.core collections.
 
 ## Example Playbook
 
-Including an example of how to use your role (for instance, with variables passed in as parameters) is always nice for users too:
+Including an example of how to use your role (for instance, with variables passed in as parameters) is always nice for users too.
+
```
branic: Let's just get rid of these two lines. I don't think they are necessary and are just from the template.

roles/aap_ocp_install/README.md
```diff
@@ -125,6 +149,38 @@ Including an example of how to use your role (for instance, with variables passe
...
+      password: <PASSWORD>
+      validate_certs: false
+    aap_ocp_install_namespace: aap-test
+    aap_ocp_install_operator:
+      channel: "stable-2.5-cluster-scoped"
+    aap_ocp_install_platform:
+      instance_name: automationcontroller
+      namespace: aap-platform
+    aap_ocp_install_controller: true
+    aap_ocp_install_eda: true
```
branic: Setting these to `true` isn't obvious in the sections above for each key.

I wonder if it wouldn't be clearer to have something like an `install` key in the dictionary that is set to true or false. So that the usage is similar across versions. The `install` key could also be added to the the `aap_ocp_install_hub` dictionary so they all functioned similarly.

roles/aap_ocp_install/README.md
```diff
@@ -20,6 +20,9 @@ A description of the settable variables for this role should go here, including
 | aap_ocp_install_operator                     | Yes*     | None          | YAML Manifest to override the generated operator `Namespace` resource                        |
 | aap_ocp_install_controller                   | Yes*     | None          | Dictionary containing keys defined in the `controller variables table`                       |
 | aap_ocp_install_hub                          | Yes*     | None          | Dictionary containing keys defined in the `hub variables table`                              |
+| aap_ocp_install_eda                          | Yes*     | None          | Dictionary containing keys defined in the `eda variables table`                              |
+| aap_ocp_install_platform                     | Yes*     | None          | Dictionary containing keys defined in the `platform variables table`                         |
+| aap_ocp_install_lightspeed                   | No       | None          | Indicates whether to install the Lightspeed component in an AAP platform installation        |
```
branic: Need to specify that this is only applicable to 2.5+ platform install and what the acceptable values are and that the default is false.

roles/aap_ocp_install/tasks/install-platform.yml
```diff
@@ -0,0 +1,172 @@
...
+    validate_certs: "{{ aap_ocp_install_connection['validate_certs'] | default(omit) }}"
+    method: GET
+    status_code:
+      - 200
+  when:
+    - aap_ocp_install_hub is defined
+  register: _aap_ocp_install_hub_available
+  until: (_aap_ocp_install_hub_available['status'] == 200) and ('migrations_notran' not in _aap_ocp_install_hub_available['url'])
+  retries: 120  # Wait for 30 minutes (120*15/60)
+  delay: 15
```
branic: This check is failing for me. Doing a get of the Hub route with `/api` returns a `status_code` of 404. However, using a path of `/api/galaxy/pulp/api/v3/` returns a 200 `status_code` and allows the task to complete.

roles/aap_ocp_install/tasks/pre-validate.yml
```diff
@@ -94,23 +94,43 @@
...
+    __aap_ocp_install_major_version: "{{ aap_ocp_install_operator['channel'] | regex_search(__aap_ocp_install_major_version_re, '\\1') | first }}"
+    __aap_ocp_install_minor_version: "{{ aap_ocp_install_operator['channel'] | regex_search(__aap_ocp_install_minor_version_re, '\\1') | first }}"
+  vars:
+    __aap_ocp_install_major_version_re: '(\d+)\.\d+'
+    __aap_ocp_install_minor_version_re: '\d+\.(\d+)'
+
+- name: Set a flag indicating whether to use the new (AAP 2.5+) operator installation method
+  ansible.builtin.set_fact:
+    __aap_ocp_install_25_install: "{{ (__aap_ocp_install_major_version | int) > 2 \
+                                  or (( __aap_ocp_install_major_version | int) == 2 \
```
branic: To fix an ansible-lint warning:

```suggestion
                                  or ((__aap_ocp_install_major_version | int) == 2 \
```

roles/aap_ocp_install/tasks/pre-validate.yml
```diff
@@ -94,23 +94,43 @@
...
+                                  and ( __aap_ocp_install_minor_version | int) >= 5) }}"
```
branic: To fix an ansible-lint warning:

```suggestion
                                  and (__aap_ocp_install_minor_version | int) >= 5) }}"
```

changelogs/fragments/add_eda_install_on_ocp.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - Added ability to install EDA Controller on OCP
```
branic: This changelog fragment and filename don't describe what change this PR is actually making.

roles/aap_ocp_install/tasks/pre-validate-platform.yml
```diff
@@ -0,0 +1,198 @@
...
+          - (aap_ocp_install_controller['projects_storage_size'] | regex_replace('^(?P<size>[0-9]+)(?P<si>Ki|K|Mi|M|Gi|G|Ti|T|Pi|P|Ei|E)$', '\\g<si>')) is in ['Ki','K','Mi','M','Gi','G','Ti','T','Pi','P','Ei','E']
+        quiet: true
+  rescue:
+    - name: Update validation errors fact - projects_storage_size replicas
+      ansible.builtin.set_fact:
+        __aap_ocp_install_prevalidate_errors: >
+          {{ __aap_ocp_install_prevalidate_errors + ["aap_ocp_install_controller['projects_storage_size'] must be a number greater than 0 with a size (e.g. 12Gi or 10000M)"] }}
+
+- name: Ensure hub storage type is valid (block)
+  when: aap_ocp_install_hub is defined
```
branic: If the intention is to default the `storage_type` to `file` (which I think is a good default). Then this when clause should only be checked when a `storage_type` is defined.

```suggestion
  when: aap_ocp_install_hub['storage_type'] is defined
```

branic (changes requested):
@derekwaters This PR is looking good. I like the changes you've made after the last review. I've got a few more comments and requests and I think this PR will be good to merge.

P.S. Sorry for not seeing this sooner and getting the review done. 
P.P.S @djdanielsson thanks for the ping.

changelogs/fragments/add_eda_install_on_ocp.yaml
derekwaters: That fragment is from a previous PR (which I happened to write). I assume the fragments get compiled into the full changelog at some point? Do the fragment files get removed? I've added a new fragment for the OCP AAP2.5 update.

branic: The fragments do get compiled into the full changelog and removed when a new version of the collection is published. Looking at the commits for this PR there are commits from the other PR that added the installation of EDA. You may need to take a look at the commit history for this PR/branch and rebase it and cleanup the commits that are included.

derekwaters: Yeah, you're right, sorry, that's my mistake. I'll get rid of this file.

roles/aap_ocp_install/tasks/install-platform.yml
derekwaters: Yes, you're right, that's the endpoint exposed through the hub route (I think I was using the one on the platform route). Now corrected.

roles/aap_ocp_install/README.md
derekwaters: I've added an install: true/false key to each of the components (controller, eda, hub), along with documentation in the readme.

roles/aap_ocp_install/README.md
```diff
@@ -89,14 +115,41 @@ If the variable is omitted the corresponding component will not be installed (e.
...
+
+> ℹ️ **NOTE**
+>
+> The namespace, instance_name and link_text values will be ignored when using the platform installation method.
+
+### aap_ocp_install_platform keys
+
+| Key Name      | Required | Default Value                    | Description                                     |
+|---------------|:--------:|----------------------------------|-------------------------------------------------|
+| instance_name | Yes      | None                             | Name of the AAP Platform instance to create     |
```
branic, line 130: Can you document the `namespace` and `link_text` options here. You can just copy/paste from one of the other sections.

branic (changes requested):
@derekwaters Thanks for all your work on this. My testing has been successful.

I see one more minor change to the readme that is needed and then I will merge.

branic (approved):
Thanks @derekwaters.

PR #264, opened 2024-12-30
Parametrize async and poll to tune them as per need

myselfgagandeep:
Parametrize async and poll for long running backup for AAP2.4+ with private automation hub

# What does this PR do?

* Updated tasks and default to parametrize async and poll to respect long running backup for AAP2.4+ with private automation hub
* Avoid timeout exceeded error on backup task level where task beyond 10000 seconds times out for larger AAP environments.
* Avoid hard coding value and generalize the play.

# How should this be tested?

Provide a value above 10000 seconds to async and poll to test a backup job on AAP.

A support case is open with Red Hat to troubleshoot this.


djdanielsson (approved):
looks good, thanks

Issue #267, opened 2025-01-20
Adapt to new API status endpoints in AAP 2.5


snagoor:
## Summary

In AAP 2.5 https://issues.redhat.com/browse/AAP-37903 introduced new API status endpoints, `aap_setup_install` role should adapt to this new change.

## Issue Type

- Bug Report

## Ansible, Collection, Docker/Podman details

```console (paste below)
ansible --version
ansible [core 2.16.12]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.12/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.12.6 (main, Sep  9 2024, 00:00:00) [GCC 14.2.1 20240801 (Red Hat 14.2.1-1)] (/usr/bin/python3)
  jinja version = 3.1.4
  libyaml = True

ansible-galaxy collection list
# /root/.ansible/collections/ansible_collections
Collection                               Version
---------------------------------------- -------
ansible.posix                            1.6.0  
ansible.windows                          2.5.0  
awx.awx                                  24.6.1 
community.crypto                         2.22.0 
community.general                        9.4.0  
community.vmware                         5.1.0  
infra.aap_utilities                      2.5.2  
kubernetes.core                          5.0.0  
nshaik.aap_collection                    1.0.18 
vmware.vmware                            1.6.0 

podman --version
podman version 5.2.2

```

- ansible installation method: one of source, pip, OS package, EE

## OS / ENVIRONMENT

AAP 2.5-7 latest setup bundle to test on RHEL 9.5

## Desired Behavior

new conditional check should be introduced 

## Actual Behavior

API status checks in https://github.com/redhat-cop/aap_utilities/blob/devel/roles/aap_setup_install/tasks/setup.yml#L71-L77 would fail as per new changes in API status endpoint introduced with https://issues.redhat.com/browse/AAP-37903

djdanielsson:
looks like there was a PR to fix this, I am closing it, if there is still more work to do please reopen it.

PR #268, opened 2025-01-20
fix API checks for AAP 2.5 version

snagoor:
# What does this PR do?

This PR fixes the new changes that was introduced in https://issues.redhat.com/browse/AAP-37903 to the API status endpoints.

# How should this be tested?

Run the latest installer of AAP 2.5-7 on any latest RHEL 8 or 9 servers.

https://github.com/redhat-cop/aap_utilities/issues/267

### DIFF
--- a/roles/aap_setup_install/tasks/setup.yml
+++ b/roles/aap_setup_install/tasks/setup.yml
@@ -49,7 +49,7 @@
 
 - name: Check Gateway API Status
   ansible.builtin.uri:
-    url: https://{{ gateway_hostname }}/api/gateway/v1/status/
+    url: https://{{ gateway_hostname }}/api/gateway/v1/status/?service_keys=true
     method: GET
     user: admin
     password: "{{ aap_setup_prep_inv_secrets.all.automationgateway_admin_password | default(aap_setup_prep_inv_vars.all.automationgateway_admin_password) }}"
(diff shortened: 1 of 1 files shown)


Tompage1994 (changes requested):
I get why this has been raised, but it kind of feels like a step backwards. It feels like we should still be testing for each component's status. There would be two ways to do this. Either traverse the new array structure, or add the query `service_keys=true` (which I feel like this the more compatible option for those not using the latest releases)


From the Jira issue: _Consumers of this API can still request the previous format with a URL query parameter "service_keys=true"._

snagoor:
@Tompage1994 thank you for the review.`service_keys=true` query param was included starting with 2.5-7 installer bundle. I am not sure how this would benefit an user who is using the older bundle installer bundle (i.e. < 2.5-7 version). This parameter wouldn't exist and API calls would fail in this case.

amasolov:
> I am not sure how this would benefit an user who is using the older bundle installer bundle (i.e. < 2.5-7 version).

@snagoor I guess you are talking about AAP deployments made with the old installer. I believe a wrong parameter would be just ignored and proceed as usual which should work

Tompage1994:
> > I am not sure how this would benefit an user who is using the older bundle installer bundle (i.e. < 2.5-7 version).
> 
> @snagoor I guess you are talking about AAP deployments made with the old installer. I believe a wrong parameter would be just ignored and proceed as usual which should work

I've just checked on my instance which is from before the change and the extra query params do not cause failure

Tompage1994 (approved):
LGTM

Issue #272, opened 2025-03-18
Incorrect logic applied to lightspeed component of AnsibleAutomationPlatform


sabre1041:
## Summary

Logic for `lightspeed` component of `AnsibleAutomationPlatform` resource is backwards resulting in the Lightspeed component being installed when it is not desired when installing AAP 2.5 on OpenShift

## Issue Type

- Bug Report

## Ansible, Collection, Docker/Podman details

`main` branch

## Desired Behavior

Logic for when to install `lightspeed` as part of `AnsibleAutomationPlatform` is correct

## Actual Behavior

Lightspeed component is installed when deploying 2.5 by default

## STEPS TO REPRODUCE

Do not specify the `aap_ocp_install_lightspeed` or specify the following to attempt to not install the `lightspeed` component

```yaml
aap_ocp_install_lightspeed:
  install: false
```

PR #273, opened 2025-03-18
Corrected lightspeed logic of AnsibleAutomationPlatform resource

sabre1041:
# What does this PR do?

Correct logic for how `lightspeed` component is managed for `AnsibleAutomationPlatform` resource

# How should this be tested?

Attempt to deploy AAP 2.5 using the default configuration. `lightspeed` component should not be installed and value of `lightspeed` component will be `true`

#272


derekwaters:
Lgtm. Thanks @sabre1041

Tompage1994 (approved):

PR #284, opened 2025-06-07
Updated OpenShift install logic to support manual approval mode

sabre1041:
# What does this PR do?

Updates the logic to handle manual approval mode for `aap_ocp_install` role

# How should this be tested?

Invoke the `aap_ocp_install` role with the following variables:

```yaml
aap_ocp_install_connection:
  host: "<OpenShift_Host>"
  api_key: <OpenShift_API_Key>
  validate_certs: false
aap_ocp_install_namespace: aap
aap_ocp_install_controller:
  instance_name: "aap1"
aap_ocp_install_operator:
  channel: 'stable-2.5'
  approval: Manual
  subscription_manifest_overrides:
    spec:
      installPlanApproval: Manual
      startingCSV: aap-operator.v2.5.0-0.1728520175
```

N/A

N/A

### DIFF
--- a/roles/aap_ocp_install/README.md
+++ b/roles/aap_ocp_install/README.md
@@ -53,10 +53,6 @@ The aap_ocp_install_platform and aap_ocp_install_lightspeed Dictionaries are onl
 
 \* If the channel indicates version 2.5 or above of AAP, then the new AAP operator platform installation method will be used.
 
-> ℹ️ **NOTE**
->
-> When `approval` is set to `Manual` the operator will be installed with `Automatic` approval and then after installation the approval will be updated to Manual.
-
 ### aap_ocp_install_controller keys
 
 | Key Name                       | Required | Default Value                           | Description                                                                                                            |
--- a/changelogs/fragments/ocp-olm-manual-approval.yml
+++ b/changelogs/fragments/ocp-olm-manual-approval.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - aap_ocp_install - Enhanced support for manual approval of OLM operators.
+...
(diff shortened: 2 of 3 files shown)


djdanielsson (changes requested):
if you rebase with the PR I just did it will pass now, but can you add a changelog calling out what you are adding here?

Tompage1994:
> if you rebase with the PR I just did it will pass now, but can you add a changelog calling out what you are adding here?

The branch was open to allow that so I just went ahead and did it

branic:
I haven't test this yet, but the way this originally worked was that when the manual approval option was set the operator would be installed using the `Automatic` method and then after install the method would be changed to `Manual`. There is a note in the readme for the role to this effect:

> ℹ️ NOTE
>
>    When approval is set to Manual the operator will be installed with Automatic approval and then after installation the approval will be updated to Manual.

It looks like you are changing how this originally functioned so the readme also needs to be updated.

branic (changes requested):
Update role readme to reflect new logic

roles/aap_ocp_install/README.md
```diff
@@ -53,10 +53,6 @@ The aap_ocp_install_platform and aap_ocp_install_lightspeed Dictionaries are onl
 
 \* If the channel indicates version 2.5 or above of AAP, then the new AAP operator platform installation method will be used.
 
-> ℹ️ **NOTE**
->
-> When `approval` is set to `Manual` the operator will be installed with `Automatic` approval and then after installation the approval will be updated to Manual.
-
```
djdanielsson, line 59: not sure what happened here but this isn't showing in the latest version now

branic: Those lines are in the current role's readme (https://github.com/redhat-cop/aap_utilities/blob/4a975c3d5474a61145ce5eed5794715f6dbaede6/roles/aap_ocp_install/README.md?plain=1#L56-L58)

image

branic (approved):
I've had a chance to test this and it looks good. Thanks for the enhancement.

djdanielsson (approved):
LGTM

PR #293, opened 2025-06-26
Allow for disabling the ConsoleLink creation

branic:
# What does this PR do?

Adds an option on the OCP install to disable creating the ConsoleLink CR. This is needed when the account being used to install AAP does not have the necessary permission to create a ConsoleLink CR.

I decided to have the default be `true` so that the current behavior would be maintained. I thought about setting it to `false` as the default, but that would be a breaking change I thought it was best to avoid.

# How should this be tested?


djdanielsson (approved):
LGTM

PR #315, opened 2026-02-26
Fix deprecation warnings and standardize when/until format

branic:
# What does this PR do?

This PR is focused on the `aap_ocp_install` role Updates when clauses so that they always return True or False in support of the changes to ansible-core 2.19 and beyond. This also standardizes all of the when and until clauses to use list style syntax and ensures consistent placement of the `when` keys on a task.

# How should this be tested?

No

N/A


djdanielsson (approved):
LGTM

PR #326, opened 2026-04-02
Updated default versions and installation methods

ericzolf:
# What does this PR do?

- Fixed readme because containerized-setup isn't a valid download type
- The default installation is now containerized of AAP 2.6 on RHEL 10, instead of RPM of AAP 2.5 on RHEL 8

# How should this be tested?

You may install AAP and see that the defaults have changed

No

n/a


branic (approved):
LGTM

PR #332, opened 2026-04-14
fix(role): handle both dict and list types for hostname variables

### LINKED ISSUE #327: Misaligned type for *_hostname variables
**Is your feature request related to a problem? Please describe.**
Currently this role expects `aap_setup_prep_inv_nodes['automationcontroller']` to be of type dict. 
 https://github.com/redhat-cop/aap_utilities/blob/03b9e3971ed05c5741fabfbd0ff6b44e531e8a47/roles/aap_setup_install/defaults/main.yml#L57
But aap_configuration_template inventory shape delivers a list.

It still works currently in most scenarios because it being overwritten in the example:
https://github.com/redhat-cop/aap_configuration_template/blob/05e42020a6e93729b114593e9906f914d05d2c0d/config/qa/aap_install.yml#L74
 

**Describe the solution you'd like**
The default should be able to handle the expected type list return, like it did before
`controller_hostname: "{{ aap_setup_prep_inv_nodes['automationcontroller'][0] }}"`

**Additional context**
The same issue happens with ah_hostname when checking availablity.
The role without overwriting defaults fails with undefined variable 

```
FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: {{ (aap_setup_prep_inv_nodes['automationcontroller'].keys() | list)[0] }}: 'list object' has no attribute 'keys'. 'list object' has no attribute 'keys'. {{ (aap_setup_prep_inv_nodes['automationcontroller'].keys() | list)[0] }}: 'list object' has no attribute 'keys'. 'list object' has no attribute 'keys'\n\nThe error appears to be in '/app/.ansible/collections/ansible_collections/infra/aap_utilities/roles/aap_setup_install/tasks/containerized.yml': line 81, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: Wait for automation controller to be running\n      ^ here\n"}
```

djdanielsson:
## Summary

- Fix `aap_setup_install` hostname defaults to accept both dict and list types for `aap_setup_prep_inv_nodes`, resolving `'list object' has no attribute 'keys'` errors

## Changes

- `controller_hostname`, `ah_hostname`, `eda_hostname`, and `gateway_hostname` defaults now use Jinja2 `is mapping` test to detect whether the inventory node entry is a dict or list
- When it's a dict (from `aap_setup_prepare`), `.keys()` is used to extract the first hostname
- When it's a list (from `aap_configuration_template`), `[0]` is used directly

Closes #327

## Test plan

- [x] `ansible-lint` passes
- [x] `pre-commit` hooks pass
- [ ] `ansible-test sanity` passes
- [ ] Verified with dict-style inventory nodes (aap_setup_prepare)
- [ ] Verified with list-style inventory nodes (aap_configuration_template)
- [x] Changelog fragment added

Made with [Cursor](https://cursor.com)

### DIFF
--- a/changelogs/fragments/fix-hostname-type-mismatch.yml
+++ b/changelogs/fragments/fix-hostname-type-mismatch.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - "aap_setup_install - handle both dict and list types for ``aap_setup_prep_inv_nodes`` hostname variables, fixing ``'list object' has no attribute 'keys'`` errors when inventory nodes are provided as a list (https://github.com/redhat-cop/aap_utilities/issues/327)."
+...
--- a/roles/aap_setup_install/defaults/main.yml
+++ b/roles/aap_setup_install/defaults/main.yml
@@ -54,7 +54,7 @@ aap_setup_inst_extra_vars: {}
 # You shouldn't need to define them again and again but they should be defined
 #
 # controller_hostname: "{{ inventory_hostname }}"
-controller_hostname: "{{ (aap_setup_prep_inv_nodes['automationcontroller'].keys() | list)[0] }}"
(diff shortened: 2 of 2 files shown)


branic (approved):
Looks good.

PR #336, opened 2026-04-14
Update pre commit

djdanielsson:
updating pre commit config


djdanielsson:
we should be able to ignore the failure here and it should clear after merge

branic (approved):
Looks good

PR #340, opened 2026-04-15
ci: fix release skip chain, add sanity, remove update_pre_commit

djdanielsson:
## Summary
- **Fix release workflow**: The `create_github_release` job (and all downstream) was being skipped because `changelog` uses `always()` to handle the `major_release_gate` skip chain, but downstream jobs didn't propagate this. Added `if: ${{ !cancelled() && needs.changelog.result == 'success' }}` to `create_github_release`.
- **Add Ansible sanity CI** using the reusable workflow from `ansible/ansible-content-actions`
- **Remove `update_pre_commit.yml`** — dependabot already handles pre-commit version bumps

## Root cause
GitHub Actions gotcha: when a job uses `always()` in its `if` to run despite a skipped dependency, all downstream jobs that depend on it also need an explicit `if` condition, otherwise they inherit the "skipped" status from the transitive dependency chain.

## Test plan
- [ ] Trigger `release_auto.yml` via `workflow_dispatch` — all jobs after changelog should now run
- [ ] Verify sanity workflow runs on PRs

Made with [Cursor](https://cursor.com)


branic (approved):
Looks good

branic:
@djdanielsson The CI requirements is broken as the `pre-commit_and_sanity / all_green` check isn't being run so the PR can't be merged.

sean-m-sullivan (approved):

PR #355, opened 2026-06-08
Remove disabled actions-cool/issues-helper workflows

### LINKED ISSUE #354: CI workflows fail daily: actions-cool/issues-helper repo disabled by GitHub
## Summary

The 4 issue-lifecycle workflows in this repo (`issue-close-inactive.yml`, `issue-find-inactive.yml`, `issue-labeled.yml`, `issue-remove-inactive.yml`) have been failing on every scheduled run since GitHub disabled the `actions-cool/issues-helper` action.

Runs error out at the "Prepare all required actions" step within ~0.4s:

```
Getting action download info
##[error]Repository access blocked
```

**Background:** GitHub disabled the `actions-cool/issues-helper` repo following the May 2026 "TeamPCP" / "Mini Shai-Hulud" supply-chain campaign. All tags and underlying commits in that repo were retargeted to attacker-controlled code, so the pin currently in these workflows (`200c78641dbf33838311e5a1e0c31bbdb92d7cf0`) does not mitigate — those refs *are* the attacker commits.

## Issue Type

- Bug Report

## Ansible, Collection, Docker/Podman details

N/A — this is a CI workflow / GitHub Actions config bug, not a runtime collection bug.

## OS / ENVIRONMENT

GitHub Actions (ubuntu-24.04 runners), affects all scheduled and event-triggered runs of the listed workflows.

## Desired Behavior

The issue-lifecycle workflows succeed (or are removed if no longer needed).

## Actual Behavior

Workflows fail at action download with `Repository access blocked`.

## Suggested Fix

Either:
1. Delete the affected workflow files entirely (each contains only the `actions-cool/issues-helper` step — no other logic would be lost).
2. Replace the step with `actions/github-script@v7` using inline JS (the close/find/label/remove patterns are ~10 lines each).

We will follow up with a draft PR taking option 1 — feel free to close it if you prefer option 2.

## Security note

We do not believe there is active secret exposure — GitHub's block on the upstream action prevents any compromised code from executing on current runs. The brief window between the action getting compromised and GitHub blocking it (early-to-mid May 2026) is the only period during which a runtime `GITHUB_TOKEN` could have leaked; default permissions on these workflows were limited to `issues:write` + `metadata:read`, so blast radius was bounded.

## Detection source

[fluxgate](https://github.com/north-echo/fluxgate) v0.7.6 scan, FG-022 rule (Known Vulnerable Action), 2026-06-08.

---

*Filed by [north-echo](https://github.com/north-echo) as part of routine supply-chain hygiene scanning across public OSS orgs.*

north-echo:
Closes #354.

## What

Removes workflow files that depend on `actions-cool/issues-helper`:

- `.github/workflows/issue-close-inactive.yml`
- `.github/workflows/issue-find-inactive.yml`
- `.github/workflows/issue-labeled.yml`
- `.github/workflows/issue-remove-inactive.yml`

## Why

GitHub disabled the upstream `actions-cool/issues-helper` repo following the May 2026 TeamPCP / Mini Shai-Hulud supply-chain campaign. Every scheduled run of these workflows currently fails within ~0.4s at the action-download step (`##[error]Repository access blocked`), so they produce only CI noise. The pinned ref `200c78641dbf33838311e5a1e0c31bbdb92d7cf0` is itself attacker-controlled, so re-pinning is not an option.

Each affected file contained only the `actions-cool/issues-helper` step (no other meaningful logic), so deleting outright is the minimal change.

## Alternative

If issue-lifecycle automation is still wanted, `actions/github-script@v7` with inline JS is a drop-in replacement (~10 lines per action). Happy to follow up with that variant if you'd prefer — just close this PR and we can.

## Verification

Sample failing run: see the issue linked above.

---

*Submitted by [north-echo](https://github.com/north-echo) as part of routine supply-chain hygiene scanning across public OSS orgs.*


north-echo:
The two failing checks (`pre-commit` and `sanity / sanity-py3.12-devel`) appear to be pre-existing on `devel`, unrelated to this diff. Root cause:

```
ERROR: Package 'ansible-core' requires a different Python: 3.12.13 not in '>=3.13'
```

`ansible-core@devel` now requires Python 3.13+, so the `sanity-py3.12-devel` matrix entry fails immediately at the `pip install` step — before any code from this PR is exercised. This PR only deletes four workflow files (`.github/workflows/issue-*.yml`); nothing it touches could cause a Python version mismatch in the sanity matrix.

Verified via `gh run list --repo redhat-cop/aap_utilities --branch devel` showing the same failure pattern on default-branch runs.

